PR #107, opened 2011-09-09
Add STAGING PARTIAL CHANGES topic to documentation

rosenfeld:
Please, take a look if you find these instructions could be useful for documenting better your last commit.

Thanks!

### DIFF
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -44,7 +44,8 @@ that are part of Git repositories).
                         ca    |Gcommit| --amend
                         D     |:Gdiff|
                         ds    |:Gsdiff|
-                        dp    |:Git!| diff (p for patch; use :Gw to apply)
+                        dp    |:Git!| diff (p for patch; use :Gw to apply - See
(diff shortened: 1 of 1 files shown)


doc/fugitive.txt
```diff
@@ -260,6 +261,23 @@ a statusline, this one matches the default when 'ruler' is set:
...
+
+With Git, you can select partial chunks of code to send to stage area with "git add -e"
+or "git add -p". See "git help add" for more information.
+
+With fugitive, you can accomplish the same in two different ways:
+
+1 - after :Gstatus, you can type "dp" in the line containing the file you want to filter
+chunks of changes.
+2 - you can just type ":Git! diff %" (or any other filename instead of '%', which means
+the file in current buffer).
```
blueyed: This should probably also reference `:Gdiff` here, which is often more useful.

tpope:
Could still use improvements I'm sure but the more elaborate mappings documentation should help with a lot of this.

Issue #112, opened 2011-10-01
:Gblame 'u' shortcut to remove current patch


ramses0:
Maybe there is a better way to do this, but I'm dealing with a project that is pretty old with lots of authors, lots of commits (especially old, big svn-ish commits, 1k+ line diffs using `git show HEAD~1000 | wc -l`).

I'll see a :Gblame'd function:

```
ff1234
ff1234 function foo() {
44abc    if ( xyz == 999 ) {
44abc       alert(1);
44abc    }
ff1234 }
ff1234
```

I know I can navigate to "44abc" and press "o" to open the diff (that might give me some context clues).  I know I can press "enter" to be taken to "the revision of the whole file as it was in 44abc" (which might be 1+ years ago).  But sometimes it "feels like" it would be helpful to be able to navigate to "44abc", press "u", and effectively:

```
git diff 44abc 44abc~1 SomeFile.js | patch && reload SomeFile.js
```

Ideally with the approximate cursor position / same function?  Maybe nearest git-hunk header reference? (http://www.robertames.com/blog.cgi/entries/git-diff-hunk-headers-function-name.html)  Maybe requiring a visual selection?

I'm trying to think of a decent algorithm that would make this happen, but nothing is coming right away.

Basically it'd be super helpful to be able to check whether "44abc" was simply splitting the conditional alert from one line to 3, changing the triggering condition, changing from console.log to alert, etc...

Thanks...

PR #114, opened 2011-10-04
working mockup of 'p' and '!' support for preview-window blaming

ramses0:
related to github issue #112 make it more comfortable to walk through long
stretches of history in a :Gblame window by adding preview window support
via "p" and adding "!" which combines "~" with opening the preview commit.

---

code is a little ghetto since I don't trust myself to refactor vim-script but at least this is a low-friction way for you to try the idea.

repeated uses of 'p' in a :Gblame window feels pretty good, and the parity between `~` and `!` also feels pretty comfortable.


tpope:
`p` is implemented, I think I'm going to pass on `!`.

PR #175, opened 2012-03-26
Vertical split mapping from Gstatus window

asgeo1:
Please include - my monitor doesn't like horizontal splits :)


whcn:
Hello, how can I preview Gstatus in a vertical window?

Issue #206, opened 2012-05-26
feature request: changed flag in statusline


stardiviner:
I hope fugitive can give out a flag in statusline indicates that this file in Git or any VCS else is changed (different with file in Vim buffer or Disk). 
I always use fugitive to modify config file, etc, sometimes the file is long, I do not know where I modified.
So a statusline flag will notice me. I do not always go to see git diff. It waste time very much. Because usually I just modify, then commit.

xanderdunn:
An awesome extension of this would be the option to show flags for lines that have been changed within the past x days.  This would be super helpful for debugging.  I know a bug was introduced within the past x days, so now I need to figure out which lines of code caused it.  Having a gutter indicator on the lines that were changed would be amazingly helpful.

blueyed:
See https://github.com/tpope/vim-fugitive/pull/219 for an older PR in this regard.

@stardiviner 
Why have you closed the issue?

stardiviner:
@blueyed sorry, I thought this feature was implemented.

zhiyuanzhai:
Is there any update?

tpope:
My update is that if people are going to nag me about dormant but not abandoned feature requests in the issue tracker then I guess I will have to close them.

Issue #197, opened 2012-04-26
Feature Request: change status line if files have been modified


jgallen23:
It would be nice to see, at a glance, if I have modified files in my git directory.  It could be as simple as [Git(master)*] if something has been modified

tpope:
Filing this under #206 (although boo on them for opening a dupe one measly month later

PR #219, opened 2012-07-03
Show status flag of repo in statusline

elien122:
This is customizing the fugitive statusline for showing status flags which indicate current repository status.
[like $(__git_ps1) of git-completion.bash in linux shell.](http://en.newinstance.it/2010/05/23/git-autocompletion-and-enhanced-bash-prompt/)

I wrote this tweak a couple of month ago just for my taste.
By chance i stopped by your github site, and i saw someone who needs this feature.
(issue #206, #197)

Tested on WinXP-x86, Ubuntu-2.6.32.32-x64. Vim version on both platform is 7.3.566
I hope this will help.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -12,6 +12,23 @@ if !exists('g:fugitive_git_executable')
   let g:fugitive_git_executable = 'git'
 endif
 
+
+"initialize variables for repo_status_flag on statusline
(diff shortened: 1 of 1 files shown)


tpope:
If this does happen, it would require an all new implementation at this point, 7 years in the future. Let's make #206 the umbrella issue.

PR #276, opened 2012-12-02
Misc changes for fugitive

ekacnet:
Hello, 

I did a couple of changes to fugitive, mainly to support notes in fugitive (both displaying and editing) and also added a new mapping to be able to go back and forth in changes.


tpope:
If anything here is still relevant, please feel free to resubmit, one feature per PR.

PR #279, opened 2012-12-09
Add :GBrowser support for Gitlab Site

yangchenyun:
I am not familiar with vim script and vim doc, so if there is anything not fit in the style, let me know and I will fix it.

## Why?
gitlab use `/tree/` instead of `/blob/` in the url to view code in the browser.

## Solutions
added a global switches for that
it works together with the `g:fugitive_github_domains` to support gitlab sites

## Examples
With the following setup:

    let g:fugitive_github_domains = ['http://git.zenhacks.org']
    let g:fugitive_gitlab_engine = '1'

I could use `:GBrowser` seamlessly with gitlab.

### DIFF
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -200,10 +200,20 @@ that are part of Git repositories).
                         "git instaweb" from a terminal).  If a range is given,
                         it is appropriately appended to the URL as an anchor.
 
-                        To use with GitHub FI, point g:fugitive_github_domains
+                        It also supports GitHub FI and Gitlab Site.
+
+                        To use them with fugitive, point g:fugitive_github_domains
                         at a list of domains:
 >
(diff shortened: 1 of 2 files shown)


linduxed:
This PR seems to have gone completely stale, however the functionality is interesting.

What's required for this PR to get merged? Can it be merged?

tpope:
I exposed an API for this. See #445, and https://github.com/tommcdo/vim-fubitive for an example implementation.

shumphrey:
In case anyone searches and finds this closed issue, https://github.com/shumphrey/fugitive-gitlab.vim implements gitlab support using the fugitive API

rkrdo:
@shumphrey Thank you for that! I love this feature and was bummed that I couldn't use it with gitlab.

0x62ash:
Thanks for plugin.

Issue #318, opened 2013-03-14
Easy way to open diffs from :Gstatus in a new window


qstrahl:
The diff commands in the :Gstatus window (`D`, `ds`, `dv`) all go to the last active window (or stay in the :Gstatus window if no other windows exist) and start the diff there, leaving me with the chore of navigating back to the file formerly in that window once I'm done. Does there exist a way to make these commands open a new window (or tab, maybe, depending on user preference?) so as to avoid messing up my layout? If not, can that be done?

acornejo:
+1 for diff in new tab, happy to implement this as a pull-request if it has a chance of being accepted.

tpope:
Here's a crazy idea: what if using the diff maps in visual mode opened a diff for each file in its own tab? I think this is the only sensible behavior for diffing multiple files at once, so it's unlikely we'll ever want the visual mode maps for something else, and it scales down to the one file case just fine.

StrongerXi:
Any update on this? Would be a really helpful feature imo.

PR #319, opened 2013-03-18
Add :Gtdiff, :Gtsdiff, :Gtvdiff, and related mappings

qstrahl:
Upon further reflection,  I concluded that diffs in a new split are a little too silly. Diffs in a new tab, though, still seem like a good idea, so I added it. This closes issue #318
- `:Gtdiff` is `:Gdiff` but in a new tab
- `:Gtsdiff` is `:Gsdiff` but in a new tab
- `:Gtvdiff` is identical to `:Gtdiff` (for symmetry with `:Gtsdiff`)
- `dt` in the status window does `:Gtdiff`
- `dts` in the status window does `:Gtsdiff`
- `dtv` in the status window does `:Gtvdiff`

### DIFF
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -49,6 +49,9 @@ that are part of Git repositories).
                         dp    |:Git!| diff (p for patch; use :Gw to apply)
                         dp    |:Git| add --intent-to-add (untracked files)
                         dv    |:Gvdiff|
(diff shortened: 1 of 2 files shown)


blueyed:
See https://github.com/tpope/vim-fugitive/pull/413 for a follow-up.

Issue #76, opened 2011-05-15
New option to start Gdiff in a new tab


ronniel1:
Hi,

   I really like this plugin.
One thing that would make it easier to use, for me, is a variant of Gdiff to open the diff in a new tab.

Thanks
Ronnie

acornejo:
Can this solution be easily extended to open diff in a tab while in the Gstatus window?

The obvious thing which is

```autocmd FileType gitcommit noremap <buffer> dt :GdiffInTab<CR>```

Doesn't really work for various reasons (new tab looes Gstatus state, and we would want to open a tab and then do Gdiff with the file on the current line of Gstatus). This really feels like something that would be better if built into fugitive.

**Comment by acornejo** [comment]
For posterity this is the hack I came up with

```
function GStatusTabDiff()
  if has('multi_byte_encoding')
    let colon = '\%(:\|\%uff1a\)'
  else
    let colon = ':'
  endif
  let filename = matchstr(matchstr(getline(line('.')),'^#\t\zs.\{-\}\ze\%( ([^()[:digit:]]\+)\)\=$'), colon.' *\zs.*')
  execute ':Gedit ' . filename
  Gvdiff
endfunction
command GStatusTabDiff call GStatusTabDiff()
autocmd FileType gitcommit noremap <buffer> dt :GStatusTabDiff<CR>
```

The regexps look pretty complex, but I decided to copy them from what fugitive uses internally since its battle tested.

sflip:
While @acornejo was helpful for me, it doesn't seem to work anymore in recent versions of this plugin. See [my answer here](https://github.com/tpope/vim-fugitive/issues/1451#issuecomment-765576235) for a working alternative.

Issue #36, opened 2011-02-03
A noob question, exiting out of Gdiff


wavded:
Love this plugin, but one thing I can't figure out is how to stop doing a Diff.  If I use:
    :Gdiff
It will split my windows out and let me do that, but how to do I exit that.  I've tried:
    :diffoff
But that doesn't get me back to my original file and remove the other "orig" version.  Any tips?  Thanks.

Marc

starikovs:
The way I do a diff:

1. :Glog --
2. :cw to open a list of commits
3. Navigate to a commit and its info as well affected files are shown
4. Focus on a file path
5. <c-w>gf and the diff is opened in new tab
6. :tabclose to just close the tab and get the previous state before the diff

https://stackoverflow.com/a/45075152/202550

tandrewnichols:
> Yeah no problem I'll just store them all on the SPACE SHUTTLE THAT VIM ALSO SHIPS WITH.

How do I access said space ship? `:h spaceship` is missing. Please advise.

PR #413, opened 2013-12-06
add 'diff in new tab'

yssl:
I've already searched and read postings related with "diff in tab page" topic ([Issue #36](https://github.com/tpope/vim-fugitive/issues/36), [Issue #76](https://github.com/tpope/vim-fugitive/issues/76), [Pull Request #319](https://github.com/tpope/vim-fugitive/pull/319)).

I don't think that you prefer doing diff in a tab page.
But I think not a few people want to do diff in that way, because diff in a tab page has several advantages :
- Maximized diff window would be helpful for comparing two files, especially for long files. Even for short files, one can compare two files at a first glance.
- Diff in that way does not have any influence on the original working buffers / windows in vim. After diff, one can return to their original workspace just by closing the diff tab page.

I added only two features, so I think that the commands are not getting messy and still clear.
- 'dt' in Gstatus window
- :Gtdiff command

This is my first pull request, so please understand me if I make some mistakes.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -787,7 +787,9 @@ function! s:StageDiff(diff) abort
 endfunction
 
 function! s:StageDiffTab(diff) abort
-  execute 'tabedit %'
+  let curfile = expand('%:p')
+  execute 'tabnew'
+  execute 'edit' curfile
   return s:StageDiff(a:diff)
 endfunction
 
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -786,6 +786,11 @@ function! s:StageDiff(diff) abort
   endif
 endfunction
 
+function! s:StageDiffTab(diff) abort
+  execute 'tabedit %'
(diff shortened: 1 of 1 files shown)


blueyed:
I've resolved the merge conflict, and added the following:
"Use 'tab split' also with Gtdiff; use tab before current one" (https://github.com/tpope/vim-fugitive/commit/2e3b974a60ecc30d9ac8b3a3303bd87a7afe6b51).

The branch is available at https://github.com/tpope/vim-fugitive/compare/tpope:master...blueyed:gtdiff

blueyed:
This would need documentation (see #319).

And it might be simplified as with https://github.com/tpope/vim-fugitive/pull/319, not defining an extra function.

wadkar:
So the only thing stopping this PR from merging is documentation?

wadkar:
Did I miss something? Or is this feature never going to make it?

tpope:
It doesn't follow the contribution guidelines. #715 seems to closing in favor of that.

PR #429, opened 2014-01-19
Gdiff: restore cursor position in original window

### LINKED ISSUE #428: Gdiff: Cursor position moves to top after last commit ("Focus diff window on :Gdiff")
The last change (546a6bf2197529ed1bd402244cbf6e72ed62b2ca) causes the cursor position to be on the top of the diff view, while previously it stayed at the current position.

What I mean is that if you where at the end of a file, `:Gdiff` would open the diff view, and the cursor was left there.

That has changed now, as if you would type `gg` to go to the top of the file.

This is likely caused by using `:diffsplit` now.

I have tested it with `vim -u NONE` and only fugitive being sourced.

I consider this to be a regression, because you are losing the current context.

I could imagine using an existing auto-mark (could not find a matching one), or manually storing the position before calling `:diffsplit` and jumping to it in the alternate buffer afterwards.

/cc: issue #36

blueyed:
Fixes #428

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1502,11 +1502,14 @@ function! s:Diff(bang,...)
     let commit = matchstr(spec,'\C[^:/]//\zs\x\+')
     let restore = s:diff_restore()
     let w:fugitive_diff_restore = restore
+    let curpos = getpos('.')
     if s:buffer().compare_age(commit) < 0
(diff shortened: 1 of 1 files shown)


tpope:
I've ripped out all the hacks, as they are of dubious value on modern Vim.

Issue #132, opened 2011-12-09
Easy way to move between diffs (Unimpaired-style)


Peeja:
When I'm staging a commit, I usually open `:Gstatus` and run down the diffs (with `D`) staging things.  Moving to the next diff is a complicated procedure, something like `^WP^ND`.  It feels ripe for an Unimpaired-style binding, such as `]g` to go to the next diff, and `[g` to go to the previous.  Looking at fugitive.vim, though, I can't decide how best to implement such a thing.  Unlike the Quickfix List, for instance, there's no notion of a "cursor" or "current file" in the `:Gstatus` window except for the actual cursor, which only exists when the buffer is open, and is ambiguous if it's open in multiple windows.

Any thoughts?

tpope:
Prompted by an email I've started to think about this anew. My unmerged files use case has since been addressed by `:Gmerge`/`:Gpull`, so we can set that aside. I am thinking a wrapper around `git diff --name-status` is probably the way to go. Load the results into the quickfix list, with the second file jammed into the description text, and then change `:Gdiff` to check the quickfix list for its default. You could build a higher level cycling abstraction on top of that pretty easily, though I'm not sure that's a good candidate for inclusion. For now, let's focus on the options available for populating the quickfix list.

Between no argument, `--cached, `HEAD`, and `commit1..commit2` - plus a subset of additional `git diff` arguments such as `--find-renames` - this covers a ton of use cases. It doesn't handle untracked files, which I guess isn't the end of the world. Anything else we're failing to cover?

mattboehm:
So as I understand it, the steps for the original use case would be:

* run `:Gcompare` (or whatever the alias is called)
* quickfix is populated with the changed files. The file component is a fugitive url including the commit1, the description includes commit2
    * Is the first file in the list loaded or not?
* open the first file and run `:Gdiff`
* Gdiff checks the quickfix list, realizes it's in a particular format, uses commit1/commit2 from the entry for diffing
* when done, run `:cn` then `Gdiff` to view the next

[...]

PR #469, opened 2014-04-08
Add :Glsfiles, :Gllsfiles commands

mattboehm:
These put the output of git-ls-files in the error/location list.

Note that they currently assume that the output is just the file names,
so running `:Glsfiles -s` will not work properly.

Feel free to kill this if you do not think it is a useful enough feature or to tell me to make a bunch of changes to it; you're not going to hurt my feelings. I also don't mind if you take a while to get back to me, especially since I think we agree that this is a fairly low-priority issue. 

refs #132

### DIFF
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -88,6 +88,16 @@ that are part of Git repositories).
 :Gllog [args]           Like |:Glog|, but use the location list instead of the
                         quickfix list.
 
+                                                *fugitive-:Glsfiles*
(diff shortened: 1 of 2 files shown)


tpope:
Fugitive's internals now have a feature where if you define `s:LsFilesSubcommand(...)`, it becomes available as `:Git ls-files`. I think that's where the lower value wrappers should live, rather than having a million Vim commands. This would be a good test case, if you want to have a go at it.

PR #500, opened 2014-06-12
Add the :Gtag command for tag creation

idanarye:
:Gtag is for tag creation only - tag listing and tag deletion is
blocked. Since these commands don't require a message, they should be
used with :Git.

The tag message buffer uses the gitcommit filetype since any ftplugin
that helps in writing commit messages and targets the gitcommit buffer
will probably be useful for writing tag messages as well.

### DIFF
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -70,6 +70,13 @@ that are part of Git repositories).
                         git-add and git-reset while a commit message is
                         pending.
 
+                                                *fugitive-:Gtag*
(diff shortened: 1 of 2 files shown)


tpope:
I had overlooked this when I closed #501. Very open to it but I'm afraid you'd basically be starting over at this point. I am fine with a private function that's super narrowly targeted.

tpope:
Try the `job` branch. It has generic support for editor invoking commands which should hopefully solve this once and for all.

PR #501, opened 2014-06-13
Add fugitive#execute_with_message_buffer()

idanarye:
Also see #500

fugitive#execute_with_message_buffer is a framework function that
generalizes what s:Commit does. It runs a git command, and if that
command causes Git to open an editor it uses a Vim buffer instead.

fugitive#execute_with_message_buffer receives a single argument which is
a Vim dictionary with all the other arguments:
- 'git_command': The base git command to run
- 'msgfile': The name of the file inside the .git folder that holds the
  message that git would open an editor to edit.
- 'args': The command line arguments that go after the git_command.
- 'sanitize_boolean_flags': A list of boolean flags that should be
  removed from the command line arguments in case a buffer needs to be
  opened.
- 'sanitize_boolean_flags': A list of value accepting flags that should
  be removed from the command line arguments in case a buffer needs to
  be opened.
- 'post_command_callback': A function that will be called after the
  command was successfully ran.
- 'msg_buffer_opened_callback': A function that will be called after a
  buffer was opened for editing the message.

Notice I've also removed the `fugitive_commit` `augroup`, since that buffer can now have names other than 'COMMIT_EDITMSG'. Instead the `aucmd` is created with the `<buffer>` flag directly inside the `fugitive#execute_with_message_buffer`.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -939,14 +939,31 @@ function! s:StagePatch(lnum1,lnum2) abort
 endfunction
 
 " }}}1
-" Gcommit {{{1
+" fugitive#execute_with_message_buffer {{{1
+function! s:RegexForMatchingFlags(flags)
+  let l:single_char = ''
+  let l:multi_char = []
+  for l:flag in a:flags
+    if 1 == len(l:flag)
+      let l:single_char.=l:flag
+    else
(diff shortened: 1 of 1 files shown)


tpope:
Thanks for taking a swing at this. `:Gmerge` had some different requirements so I ended up going a different direction.

Issue #121, opened 2011-10-26
:GDiff broken?


teranex:
When I try to run :GDiff on a file which has changes I get the following error:

Error detected while processing function <SNR>21_Diff:
line   44:
E480: No match: `=spec`  

I'm not really sure what is causing this. I checked out Fugitive version 1.2 and 1.0 but they give me the same results.
I'm pretty sure I have used this feature before.

I'm on Xubuntu 11.10 with Gvim 7.3 with patches 1-154 (default vim for 11.10)

PR #532, opened 2014-08-06
Work around 'wildignore' conflicts.

justinmk:
Use s:fnameescape instead of backtick expressions.

References #121.

I removed `.git` from `wildignore` long ago, but IT recently changed my environment and I'm getting errors related to this again, presumably because of some other thing in `wildignore`. 

My perspective: If wildignore is of any use, one shouldn't have to worry that it's going to break plugins. I use `wildignore` for `:vimgrep` and `find()`, but this user-facing option should not affect the internal operation of plugins, for the same reason that plugins call `:norm!` instead of `:norm`.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -349,10 +349,10 @@ function! s:repo_git_chomp_in_tree(...) dict abort
   let cd = exists('*haslocaldir') && haslocaldir() ? 'lcd ' : 'cd '
   let dir = getcwd()
   try
-    execute cd.'`=s:repo().tree()`'
+    execute cd.s:fnameescape(s:repo().tree())
     return call(s:repo().git_chomp, a:000, s:repo())
   finally
-    execute cd.'`=dir`'
+    execute cd.s:fnameescape(dir)
   endtry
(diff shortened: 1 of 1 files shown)


talha131:
@justinmk Any update on this?

justinmk:
@talha131 I gave up on this. Instead I set only a very minimal wildignore (i.e. things I don't ever expect to use vim for). Definitely don't put `.git` in 'wildignore'.

The reason is that even if one fixes this for fugitive (personally I still think it should be), certain features of vim like the "backtick expressions" will be broken. 

I agree with @tpope that 'wildignore' _should not_ be used as a general mechanism as is done by plugins like ctrlp. 'wildignore' is too low-level for such purposes. On the other hand as a plugin author it bothers me to think about all the users that are seeing silently weird/broken behavior in plugins like fugitive just because of a misconfigured 'wildignore'.

I'm going to add a warning in Neovim's `:CheckHealth` that alerts users about this.

tpope:
FYI there are definitely some syntax problems in that pull request. I would entertain a corrected version.

PR #540, opened 2014-08-25
use more exact time regex for Gblame

lilydjwg:
Sometimes author names ending with digits and get highlighted wrongly.


tpope:
Git supports more date formats than `YYYY-mm-dd`, so this is insufficient. But I'm happy to brainstorm if you or someone provides examples that fail.

PR #542, opened 2014-08-27
Feature: Add command to vertically split a commit on Gblame

WooterTheTroubleshooter:
Add key S to Gblame for a vertical split of the commit

S is also used for split on Gstatus
(Edit: I ment vertical split of the commit ;))


tpope:
This is on `gO` now.

PR #562, opened 2014-10-20
Add :Gwn, :Gwp, and :GwN commands

glittershark:
The way I've done it these are literally copy-pastes of the function and command definitions for `Gwq`, so this is probably a candidate for DRYing out. Plus there's no documentation; I'm happy to add that if you like - let me know if you'd like that in a seperate commit or in this one.


glittershark:
*bump*

lieryan:
This looks like a fairly simple mapping to add to your .vimrc:

    command Gwn :Gw|next
    command Gwp :Gw|prev

PR #583, opened 2014-12-02
Expand `%` instead of `<amatch>` in the Buf* autocommands

blueyed:
This is relevant in case a previous BufReadPost autocmd changes the file
name using `:file` - fugitive should use the new/current name then.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -218,7 +218,7 @@ endfunction
 
 augroup fugitive
   autocmd!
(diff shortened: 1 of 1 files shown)


blueyed:
This seemed to be an obvious fix by then, please review / provide feedback, or even merge it.. :)

blueyed:
Closed in https://github.com/tpope/vim-fugitive/commit/0b43b51d7785aeb4002b45ca49cea5aef0d2e988. Thanks.

Issue #421, opened 2014-01-03
WinEnter/BufEnter autocommand not triggered when closing "fugitive://" buffer after :Gdiff


blueyed:
My WinEnter/BufEnter autocommand is not being triggered when closing (`:q`) the "fugitive://" buffer after `:Gdiff`.

When the non-fugitive:// buffer is closed, the BufEnter autocommand is triggered.
It is also triggered after a normal `:sp`, followed by `:q`.

Is this behaviour expected?
Any hints how to debug this?

blueyed:
Fixed in https://github.com/tpope/vim-fugitive/commit/e065e4f841f4a033ac9a6156573a446007617b1d.

Thanks @tpope!

PR #585, opened 2014-12-11
(Re)store the current window in `s:diffoff_all`

blueyed:
This is required to make Vim execute the "Enter" autocommands when
closing the fugitive window.

Fixes: https://github.com/tpope/vim-fugitive/issues/421

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1667,6 +1667,7 @@ function! s:diffoff() abort
 endfunction
 
 function! s:diffoff_all(dir) abort
+  let curwin = winnr()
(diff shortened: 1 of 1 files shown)


blueyed:
@tpope 
Any objections on this fix?

Because it fixes/restores Vim's autocommand behavior this is likely to fix other oddities, too.

blueyed:
@tpope 
I'm sorry to bug you again, but this issue causes unexpected behaviour when using `:Gdiff`, and the fix appears to be trivial.

tpope:
Okay, guess I'm fine with merging blind.

Issue #435, opened 2014-02-10
Can't run :Gcommit on Windows 7 with PowerShell


tbrosnan:
I'm seeing the following error message "Can't open file C:/.../Appdata/Local/Temp/VILA140.tmp". That temporary file doesn't appear to exist.

I am using GVim 7.4, running :Git --version shows "git version 1.8.3.msysgit.0".

justinmk:
Does anyone have `set shell=powershell` working with Vim+fugitive at this point?

@Konfekt `let &shellpipe='| Out-File -Encoding UTF8 %s'` seems to break things like `:r!ls` (which works with `shellpipe=| shellredir=>`).

dragon788:
@justinmk I've been having many failures with Vim+PowerShell as my working combo, which makes me really sad as I need to work in PowerShell for work, and I'd prefer to use Vim to do so, but these failures mean I probably need to use PowerShell and fall back to cmd.exe as the Vim shell and pass /c powershell -Command to do anything. :-1:

justinmk:
@dragon788 I use `co!` mapping to toggle powershell on/off: 

https://github.com/justinmk/config/commit/bb1469f870f0a9a8613802513ff5acdf1e1e19fe#diff-4e12c6a37ff2cbb2c93d1b33324a6051R493

Yes, this sucks.

dragon788:
Interestingly, I seem to have less issues with it failing once I disabled
the solarized plugin. I was seeing these failures when using vim plug to
install or update plugins in Windows.

At first I thought it was some of my symlinks causing issues, but I went
through and unmade and remade them and was able to reproduce the issue in
all my various iterations, but as soon as I removed solarized and cleared
out my plugins it worked fine.

I may need to switch to another theme and see if it is a particular issue
there or if it is reproducible depending on the plug-in list.

I did set up the mapping and it does seem to help in some cases, so I'll
have to see about testing both ways.

Apologies Tim, I do like a pathogen but experiment from time to time.

Konfekt:
There is vim-flattened as a stripped down alternative to solarized.

tpope:
How well does PowerShell play with Vim jobs? Been thinking of using jobs to make `:Gstatus` snappier and could see rolling it out for everything if it works out well.

Konfekt:
[...]

PR #586, opened 2014-12-14
use same escaping for cmd and powershell

Konfekt:
Powershell, just like cmd, uses doubled double quotation marks 
to escape double quotation marks. Therefore s:shellesc() should 
take account of cmd and powershell as well.

This was discussed at https://github.com/tpope/vim-fugitive/issues/435


sinisterstuf:
Is this a wanted feature? After almost 3 years chilling here, I guess not really. Maybe it and its related issue could be closed?

tpope:
Assuming unwanted until someone says otherwise!

PR #601, opened 2015-02-02
[WIP] Add range support to :Gwrite

### LINKED ISSUE #370: Feature request: :Gwrite should accept a range
I use vim-signify to see signs corresponding to `git diff` output.  Because of this, I already know where diff hunks exist, and so staging them directly from the buffer with a single command would be great (rather than dropping into a split diff view, `:diffput`-ing, and closing the diff view).  Allowing `:Gwrite` to accept a range for its operation would be great; presumably, a higher-order `:Ghunkw` (or whatever) command could then be cobbled together that used vim-signify signs' line numbers to determine the range to be provided to `:Gwrite`.

sharplet:
This is an attempt to resolve #370.

I took a look at [GitX](https://github.com/pieter/gitx), an app I use that does support this feature. [Its
implementation](https://github.com/pieter/gitx/blob/a256210880f2878707f1714515663f5f09360df8/html/views/commit/commit.js#L273-L331) is roughly as follows:
- Get the selected lines
- Rewrite the patch to include only the selected changes
- Apply the rewritten patch

I'm wondering if a useful first step would be to first implementing working with
hunks (i.e., have a command/binding to add/reset just the hunk under the
cursor)? Having the ability to extract just the current hunk may then allow the
editing of the hunk to be a bit simpler. (I'd imagine that trying to apply a
selection that spans multiple hunks would be an error.)

---

Regarding implementation: I kind of feel like it might be easier to maintain
the functionality if it was a separate command that could be used as a text
filter, i.e., send it a patch on stdin and ask it to either extract a hunk, or
format a new sub-patch by specifying a range. Just an idea though; I have no
idea whether distributing a shell script with vim-fugitive is feasible or
necessary, and maybe it would be too slow.

Perhaps then it's just a couple of new functions that would be necessary: one to
select a single hunk out of a buffer, and another to edit the hunk so it only
applies the changes within a certain line range.

---

Currently I've just added a default range parameter to `s:Write()` and `s:Wq()`
that doesn't actually do anything.

I'd appreciate any feedback you can give!

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1398,11 +1398,11 @@ call s:command("-bar -bang -nargs=* -count -complete=customlist,s:EditRunComplet
 
 " Section: Gwrite, Gwq
 
-call s:command("-bar -bang -nargs=* -complete=customlist,s:EditComplete Gwrite :execute s:Write(<bang>0,<f-args>)")
-call s:command("-bar -bang -nargs=* -complete=customlist,s:EditComplete Gw :execute s:Write(<bang>0,<f-args>)")
-call s:command("-bar -bang -nargs=* -complete=customlist,s:EditComplete Gwq :execute s:Wq(<bang>0,<f-args>)")
+call s:command("-bar -bang -nargs=* -range=% -complete=customlist,s:EditComplete Gwrite :execute s:Write(<line1>,<line2>,<bang>0,<f-args>)")
+call s:command("-bar -bang -nargs=* -range=% -complete=customlist,s:EditComplete Gw :execute s:Write(<line1>,<line2>,<bang>0,<f-args>)")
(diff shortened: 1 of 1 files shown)


tpope:
So in addition to the complexities of distributing an external script, it opens up a bunch of portability issues, so it's a nonstarter in my book. You'd have to do a pure VimL parsing of the patch. I do something sort of similar to jump to the correct line when you hit enter in `:Gblame`. My biggest concern here is the numerous edge cases around selecting half of a hunk and whatnot.

justinmk:
@sharplet `git gui` has a reliable implementation of this... in tcl. May be useful.

sharplet:
@justinmk Thanks for the heads-up. This looks promising: https://github.com/git/git/blob/master/git-gui/lib/diff.tcl#L623-L815

jreybert:
Actually, vimagit implements it in plain VimL. The main difference is that it is based on plain git diff output, which could be hard to retrieve from a Gread or Gdiff buffer...

https://github.com/jreybert/vimagit/blob/master/plugin/magit.vim#L905

Issue #394, opened 2013-10-25
No option to not map y<C-G>


dlip:
Around line 159 you have a mapping which is messing with my y key which i have mapped to 'w'. If I press y to go forward a word it hangs waiting for 2nd key. Could you please add an option to not map this key?

`nnoremap <buffer> <silent> y<C-G> :call setreg(v:register, <SID>recall())<CR>`

PR #604, opened 2015-02-11
Provide g:fugitive_no_maps to disable key maps

gusevfe:
Add variable g:fugitive_no_maps. If set, `y<C-G>` and `<C-R><C-G>` are not
mapped.

Resolves tpope/vim-fugitive#394

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -193,8 +193,10 @@ function! fugitive#detect(path) abort
         let &mls = save_mls
       endtry
     endif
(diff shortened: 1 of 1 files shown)


lobocode:
A note: vim-fugitive crashes when the system is installed the ruby, ruby-irb, rubygem-bigdecimal, rubygem-io-console, rubygem-json rubygem-psych, rubygem-rdoc, rubygems, rubypick. I don't know why, but it just does not work (tested on Fedora 21 workstation).

tpope:
That's not related to the pull request at hand, is it?

Issue #594, opened 2015-01-15
"E117: Unknown function: netrw#NetrwBrowseX" in Vim's master/tip


blueyed:
`netrw#NetrwBrowseX` has been renamed to `netrw#BrowseX` in https://github.com/vim-jp/vim/commit/9e29e82.

(around Vim's patch 567):

    -fun! netrw#NetrwBrowseX(fname,remote)
    -"  call Dfunc("NetrwBrowseX(fname<".a:fname."> remote=".a:remote.")")
    +fun! netrw#BrowseX(fname,remote)
    +"  call Dfunc("netrw#BrowseX(fname<".a:fname."> remote=".a:remote.")")

tpope:
I guess the solution here is to check for both, with the annoying caveat that that requires ensuring `autoload/netrw.vim` has been loaded.

harriott:
Is this why `:Gbrowse` ain't working for me on my Arch install (vim v7.4, patched to 580)?
If so, anything I can do about it locally, or just wait for a fix?

tpope:
You can define a `:Browse` command.

Keithbsmiley:
Here's a possible patch for this:

```diff
diff --git i/plugin/fugitive.vim w/plugin/fugitive.vim
index 65e6665..b61e49c 100644
--- i/plugin/fugitive.vim
+++ w/plugin/fugitive.vim
@@ -2225,7 +2225,11 @@ function! s:Browse(bang,line1,count,...) abort
     elseif exists(':Browse') == 2
       return 'echomsg '.string(url).'|Browse '.url
     else
-      return 'echomsg '.string(url).'|call netrw#NetrwBrowseX('.string(url).', 0)'
+      if has("patch-7.4.567")
+        return 'echomsg '.string(url).'|call netrw#BrowseX('.string(url).', 0)'
+      else
+        return 'echomsg '.string(url).'|call netrw#NetrwBrowseX('.string(url).', 0)'
+      endif
     endif
   catch /^fugitive:/
     return 'echoerr v:errmsg'
```

Seems like it will have to be a switch on the patch version since you can't check existence on an autoloaded function.

Keithbsmiley:
Until something like this is added I wrapped this into a function:

```vim
if has("patch-7.4.567")
  function! s:Browse(args)
    execute 'call netrw#BrowseX('.string(a:args).', 0)'
  endfunction

  command! -nargs=* Browse call <SID>Browse(<q-args>)
endif
```

[...]

PR #608, opened 2015-02-20
Support browsing with new netrw.vim

elyscape:
The previous fix for #594 didn't handle the case in which the net netrw was installed on older versions of Vim.


tpope:
Thanks

PR #609, opened 2015-02-21
Show all untracked files in :Gstatus window

### LINKED ISSUE #605: Gstatus fails to launch a diff if a new folder is added
If you've created new files in a directory not currently tracked by Git, `git status` by default only outputs the name of the new directory.  If you select an untracked directory in the Gstatus window and hit `D`, Gdiff gets passed the directory name.  As a result, the "current" buffer is passed that directory to open and displays a directory listing, which is not useful for a diff.

A workaround for this is to set `status.showUntrackedFiles` to `all` in one's gitconfig.  That results in `git status` enumerating all untracked files, even if they're in untracked directories.  I don't know how many versions back you support, but this feature [was added in Git v1.6.0](https://github.com/git/git/commit/d6293d1f2cadcdf45f8ada7847a32b51fa5bf0a6) and requires the user to set it.  A way for fugitive.vim to handle this itself would be to be to add the `-u` flag to its `git status` calls. The `-u`/`--untracked-files` flag [was added to `git-status` in Git v1.4.0](https://github.com/git/git/commit/443f8338b9e248353a7095a1096684f1ed106c66) and results in the full enumeration of untracked files, which solves this issue.

elyscape:
Previously, if there were untracked files inside an untracked folder,
:Gstatus would only show the new folder. Attempting to run a diff on the
folder would pass the directory name, which would result in Vim opening
a directory listing. This makes :Gstatus list all untracked files, even
if they're inside untracked folders. This requires Git >=1.4, and will
silently fall back to the old behavior on earlier versions.

Closes #605.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -2408,13 +2408,16 @@ function! s:BufReadIndex() abort
     else
       let cd = exists('*haslocaldir') && haslocaldir() ? 'lcd ' : 'cd '
       let dir = getcwd()
-      if fugitive#git_version() =~# '^0\|^1\.[1-7]\.'
+      if fugitive#git_version() =~# '^0\|^1\.[1-3]\.'
         let cmd = s:repo().git_command('status')
+      elseif fugitive#git_version() =~# '^1\.[4-7]\.'
+        let cmd = s:repo().git_command('status', '-u')
       else
         let cmd = s:repo().git_command(
               \ '-c', 'status.displayCommentPrefix=true',
               \ '-c', 'color.status=false',
               \ '-c', 'status.short=false',
+              \ '-c', 'status.showUntrackedFiles=all',
               \ 'status')
       endif
       try


elyscape:
@tpope Any chance of merging this?  Alternatively, any comments?  Sorry to bug you.

tpope:
Can't you just set `status.showUntrackedFiles` in your own config? Why should `:Gstatus` differ from `:Git status`?

elyscape:
> Can't you just set `status.showUntrackedFiles` in your own config?

Indeed, I did as a workaround, but ideally that wouldn't be necessary.

> Why should `:Gstatus` differ from `:Git status`?

So that selecting a diff on an untracked folder doesn't blow up like so:

Broken diff

Instead actually working, like so:

Functional diff

elyscape:
The reason I say that ideally setting `status.showUntrackedFiles` in my `.gitconfig` wouldn't be necessary is that I do indeed prefer the normal behavior.  Unfortunately, the normal behavior interacts poorly with running `:Gdiff` from the `:Gstatus` window.

tpope:
Oh right, skimmed over that on my reread. I guess I'll merge, though I'm not convinced the inability to diff untracked directories is that big of a tragedy.

elyscape:
Thanks!

PR #611, opened 2015-02-26
Rename ':Make' command in s:Dispatch()

AlterDepp:
The vim command ':Make' is often used to build software, so the fugitive Make command is named 'FugitiveMake'.

Change-Id: Iaf84df8c854002d6b6c0a5ae8794d2346cf45b13


tpope:
No can do, this is deliberate for interop with dispatch.vim.

PR #620, opened 2015-03-18
Add feature: Open a commit or file by double-clicking

igrep:
This feature enables us to open a file or commit by double-clicking on a window displayed by fugitive (`Gblame` and `Gstatus`).

It's implemented by copy-ing the behavior on inputting `<CR>`.


tpope:
You know ... I'm actually okay with this. I've gone ahead and redone it on master since it's a whole different paradigm. Thanks for the idea.

PR #624, opened 2015-03-28
Use the neovim terminal emulator for StagePatch()

### LINKED ISSUE #618: Gstatus "p" shortcut endless loop
This issue happens in neovim only.

When press "p" in Gstatus window, vim keeps asking for decision but  never responses to any keys. it caused vim to become unresponsive.

Tranquility:
This fixes #618

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -670,7 +670,11 @@ function! s:Git(bang, args) abort
     let git .= ' --no-pager'
   endif
   let args = matchstr(a:args,'\v\C.{-}%($|\\@<!%(\\\\)*\|)@=')
-  call s:ExecuteInTree('!'.git.' '.args)
+  if has('nvim')
+    call s:ExecuteInTree("tabnew | term ".git." ".args)
+  else
+    call s:ExecuteInTree('!'.git.' '.args)
+  endif
   call fugitive#reload_status()
   return matchstr(a:args, '\v\C\\@<!%(\\\\)*\|\zs.*')
 endfunction
@@ -708,6 +712,7 @@ augroup fugitive_status
   autocmd!
   if !has('win32')
     autocmd FocusGained,ShellCmdPost * call fugitive#reload_status()
+    autocmd BufDelete term://*:git* call fugitive#reload_status()
   endif
 augroup END
 


Tranquility:
I just realized that TabClosed was added only to neovim so another `if has('nvim')` is needed but please tell me if merging neovim specific stuff is even an option for you.

tpope:
Yeah definitely leave that out of this commit.

Is there a feature (e.g. `has('term')`) we can use instead of checking for nvim in particular?

Tranquility:
> Yeah definitely leave that out of this commit.

Do you mean I should just remove it instead of wrapping it into another if statement? This would mean the status is not updated.

> Is there a feature (e.g. has('term')) we can use instead of checking for nvim in particular?

No, because neovim is always built with all features. We might have a problem with older versions but since it's alpha state we can expect everyone to update regularly.

tpope:
Oh, I see. Can you get a buffer id back from `:term`? The closing of that buffer seems like the relevant event to target.

noahfrederick:
>> Is there a feature (e.g. has('term')) we can use instead of checking for nvim in particular?
>
> No, because neovim is always built with all features. We might have a problem with older versions but since it's alpha state we can expect everyone to update regularly.

You can check `exists(":terminal")`.

Tranquility:
I now use a BufDelete and the terminal buffer naming convention to update whenever a git command was executed in a terminal buffer.

I also added `call s:ExecuteInTree(..)` to make it work in subdirectories. It looks however like there is a race condition, when doing the `git add` the cwd gets reset to early. The `git reset` case works fine.
Do you have any idea how we can solve this?

tpope:
For starters I think you should be fixing the `:Git` command rather than this one particular special case.

Tranquility:
I decided to fix this special case because it is the only one (that I know of) that needs a tty. But I can move the fix into into the `Git` function if you prefer that.

Do you know how `git add --patch` differs from `git reset --patch` that one works while the other doesn't?

tpope:
Oh, that's really bizarre that those two commands differ. But I don't see a harm in routing all commands through `:terminal`. There are definitely other commands (such as `git rebase -i`) for which I would expect it to be necessary.

Tranquility:
I found out why they differ. Pressing <kbd>p</kbd> on a staged file executes `git reset --patch --` instead of `git reset --patch -- /path/to/file`

tpope:
Good catch. I'll definitely fix that.

Tranquility:
> I can see that this mimics Vim's behavior, which uses the whole terminal window though - so maybe there should be an option?

That's why I decided to use a tab and I felt that most hunks do not fit into a split buffer.

tpope:
Okay I've merged after the following fixes:

* Ensuring the chdir to the repo's tree root does not affect the existing buffer.
* Checking for `exists(':terminal')` rather than `has('nvim')`.
* Reloading the status after all terminal buffer deletions, because the git executable might not start with `git` (could be an absolute path, or a wrapper like hub).
* Disabling the explicit status reload except on Windows, where the autocommand is disabled.

The new tab indeed feel odd, but I'm really struggling to see a split as an improvement.

blueyed:
Just for reference, it has been merged in 4cc201cbe3a0c3faa4cde1b82ba941d410e5e81c.

sirex:
Just upgraded fugitive and was surprised to see this annoying behavior. Probably split window would make it less annoying and possibility to close splint window with one key press like ENTER or something like that, would make it behave as it was before.

If things do not fit into split window one can always move that window to another tab with `:tabedit %`, or can maximize it with `<C-W>_`.

tpope:
I agree, Neovim is annoying.

Split does sound like a better default.

PR #638, opened 2015-05-12
Add AppStream metadata.

voxik:
I am going to package fugitive.vim for Fedora and since I'd like to provide best user experience, i.e. to have this plugin shown in gnome-software [1], it be cool if you can include these metadata in your upstream repository.

[1] https://blogs.gnome.org/hughsie/2014/06/11/application-addons-in-gnome-software/

### DIFF
--- a/appstream/vim-fugitive.metainfo.xml
+++ b/appstream/vim-fugitive.metainfo.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="addon">
+<id>vim-fugitive</id>
+<extends>gvim.desktop</extends>
+<name>fugitive.vim</name>
+<summary>A Git wrapper so awesome, it should be illegal</summary>
+<url type="homepage">http://www.vim.org/scripts/script.php?script_id=2975</url>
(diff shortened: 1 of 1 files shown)


tpope:
Fugitive made it into Fedora anyways, correct? Does this still offer any value?

voxik:
It did get into Fedora. And of course this still allows gnome-software, i.e. the Fedora "App store" to list Fugitive as a VIM plugin.

tpope:
I'm missing something. Does gnome-software scrape GitHub repositories? Vim.org plugins?

carlwgeorge:
I believe gnomes-software reads it from the RPM package in the Fedora repositories.  I believe the intent of this pull request is for the metadata file to be maintained upstream (here), rather than in Fedora's packaging.  That would allow the canonical upstream file to be shared across all distributions, rather than, for example, Ubuntu and Fedora coming up with their own.

tpope:
For that scenario it's a little weird for `updatecontact` to be someone not otherwise affiliated with the project, correct? I'd presumably want to change it to myself.

I don't think I am going to add this as a one-off to Fugitive, but I will consider making it a standard practice across all my plugins. Thanks!

PR #641, opened 2015-05-16
Make Ggrep to work properly on Mac

ovu:
- Color were causing to appear strange characters in the result.
- These characters made impossible to navigate the results
- The solution was to disable colors while executing git grep


tpope:
I will merge for robustness, but you have `color.grep = always` in your `.gitconfig`, which means this is your own broken config.

PR #643, opened 2015-05-16
Force fugitive to use English.

### LINKED ISSUE #642: fugitive relies on English git
Some fugitive features rely on parsing output from git. For example, there are many instances of comparing against *Changes to be committed*. This breaks if git is another language and outputs e.g. *Cambios para hacer commit*.

I have no idea how this could be reasonably solved, but I thought I would log this issue to at least alert those who might have encountered the same problem.

Putting an alias in `.bashrc` doesn't work:

    alias git='LANG=en_US git'

I assume vim doesn't respect bash aliases.

ctford:
fugitive scrapes output from git, so commands like "-" within a :Gstatus
split fail if git's output is in another language.

Closes #642.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -9,7 +9,7 @@ endif
 let g:loaded_fugitive = 1
 
 if !exists('g:fugitive_git_executable')
-  let g:fugitive_git_executable = 'git'
+  let g:fugitive_git_executable = 'LANG=en_US git'
 endif
 
 " Section: Utility


tpope:
This only works on Bourne compatible shells.  Also it shouldn't be necessary if status hints are on.

tpope:
Can you confirm `$LANG` is set by default in your setup? We could force status hints on if `$LANG !~# '^en'`.

ctford:
Thanks for the quick response.

My `$LANG` is `es_ES.UTF-8` by default. I don't know what status hints are, so I'm not sure whether or not they're on.

tpope:
They are enabled by default and look like `(use "git add <file>..." to include in what will be committed)`.

PR #649, opened 2015-06-02
diffoff_all: restore '#' window, use winrestcmd

blueyed:
The code is based on
[`ArgsAndMore#Windo`](https://github.com/vim-scripts/ArgsAndMore/blob/848dec40e4d23fa3124c83009512eb5a22bd36b8/autoload/ArgsAndMore.vim#L127-L138).

This is a followup to e065e4f8.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1681,7 +1681,13 @@ function! s:diffoff() abort
 endfunction
 
 function! s:diffoff_all(dir) abort
-  let curwin = winnr()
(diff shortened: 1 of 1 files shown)


tpope:
@h3xx mind trying this?

blueyed:
Any feedback?

tpope:
Feel free to resubmit if you can make this work with the latest implementation (which has some `feedkeys()` hackery we might not need?).

PR #651, opened 2015-06-06
graphical log navigation

albfan:
I'm not expert on vim-fugitive internals so I prepare this quick an dirty graph log navigation

Pressing &lt;Enter> on any line shows the related commit (always in the same pseudopreview window)

Is based on AnsiEsc.vim to show colors, https://github.com/albfan/AnsiEsc.vim but ommiting "AnsiEsc" call and "--color=always" on git log command should be enough to test this (although less eye candy)

Aligning it with fugitive way to show commits, diffs, etc can be reworked if idea seems useful

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -2963,6 +2963,121 @@ function! fugitive#foldtext() abort
   return foldtext()
 endfunction
 
+function! Glogga()
+  call GlogGraph(1)
+endfunction
+
+function! Glogg()
+  call GlogGraph(0)
+endfunction
+
+function! GlogGraph(showAll)
+  silent! wincmd P
+  if !&previewwindow
+    execute 'bo ' . &previewheight . ' new'
+    set previewwindow
+  endif
+
+  execute "silent %delete_"
+  
+  setlocal buftype=nofile bufhidden=wipe noswapfile nobuflisted nowrap
+
+  let all = ""
+  if a:showAll
+    let all = " --all"
+  endif
+  let git_log_command = "git log --decorate --graph --oneline --color=always" . all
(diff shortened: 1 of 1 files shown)


albfan:
Now You can:

`:Glogg` see commits for actual branch
`Glogga` see all commits

While in log output:
`Enter` Show commit (with diffs)

While on commit output:
`Enter` Go to real from-file and num (if it exists)
`S-Enter` Go to real to-file and num (if it exists)

splits should be corrected (now divides always window in splits, but the cases where view should replace entire window must be analyzed.

tpope:
Have you thought about making this a separate plugin? I haven't looked at it closely but generally these sorts of modal windows tend to be pretty complicated, and I'm instinctively reluctant to take another in.

albfan:
Sure. Just want to know your opinion.

That was pretty much what I think you would say but I have to try.

I feel rather comfortable with that layout. 

If you ever think on extend fugitive with such pseudo-graphical interface just let me know.

albfan:
I just made the plugin. Let me know if there's a better way to integrate with fugitive than this

https://github.com/albfan/vim-fugitive-graph

tpope:
For starters, if you wanna use a Fugitive commit buffer rather than a temp file, you can `exe 'edit' fnameescape(FugitiveFind(commit))`.

albfan:
Nice, I almost added but I miss to jump directly to file in checkout if that exists. Let me open another issue (if there's no one open yet)

justinmk:
@albfan are you aware of https://github.com/junegunn/gv.vim

albfan:
@justinmk awesome. That's it. @junegunn is another beast!

Issue #243, opened 2012-08-22
Enable fugitive commands in empty buffers


chumakd:
Currently, one needs to open some file under git control in order to be able to use fugitive vim commands.

It would be great to have a possibility to run fugitive commands even if current buffer doesn't correspond to a file under git (i.e. empty buffer). It could be controlled by some vimrc option if this behavior is enabled by default or not. But of course this should work only if current PWD is inside some git repo.

PR #652, opened 2015-06-07
I killed the Beast!!!! I can now use Gcommit and Gmove in any buffer!!!

John-Appleseed:
I removed `-buffer` from augroup fugitive_remove and s:define_commands()

Tested with markdown files

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -108,7 +108,7 @@ endfunction
 
 function! s:define_commands() abort
   for command in s:commands
-    exe 'command! -buffer '.command
+    exe 'command! '.command
   endfor
 endfunction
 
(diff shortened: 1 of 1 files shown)


tpope:
You may have made them global but you certainly didn't make them work.

pencilcheck:
Is there a fix for this issue? can't run in new tab buffers

justinmk:
@pencilcheck You can

    autocmd BufWinEnter * if empty(expand('<afile>'))|call fugitive#detect(getcwd())|endif

to get fugitive to recognize a new buffer after `:tabnew`, `:new`, `:enew`.

I also do this for directory buffers (netrw or [dirvish](https://github.com/justinmk/vim-dirvish)). 

    autocmd FileType dirvish call fugitive#detect(@%)

`fugitive#detect()` seems super fast so I'm not sure why this behavior isn't the default. In https://github.com/tpope/vim-fugitive/issues/243#issuecomment-7936957 @tpope mentioned a lack of available hooks, but I wonder if the `BufWinEnter` above is acceptable?

tpope:
`fugitive#detect()` still does IO. I'm not keen on spamming it for every `BufWinEnter`.

PR #656, opened 2015-06-16
Fixes for submodules with native Vim and Cygwin Git under MSW.

vadz:
Handle "/cygdrive/x" prefix in Cygwin paths used as gitdir in .git files in
submodules when using Cygwin Git: these paths are not understood by native Vim
and its filereadable() and readfile() functions, so get rid of cygdrive prefix
to allow them to work.

---

I'm not sure if this is the best fix, but at least with these changes I can use `Gdiff` for the files in a submodule, while previously it just opened an empty split.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -43,11 +43,19 @@ endfunction
 function! s:cygfix(file) abort
   " Replace /cygdrive/x prefix with X: when using Cygwin (this prefix should
   " never be used in the path otherwise) in order to allow the path work with
-  " non-Cygwin Vim. Notice that we must capitalize the drive name for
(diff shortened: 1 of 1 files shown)


tpope:
The case insensitivity portion of this is now baked in. https://github.com/tpope/vim-fugitive/blob/c972b2a0652d284c86b509b2224fdca27d5c680b/plugin/fugitive.vim#L187 and https://github.com/tpope/vim-fugitive/blob/c972b2a0652d284c86b509b2224fdca27d5c680b/plugin/fugitive.vim#L130 are the relevant spots if you want to try the rest of your patch again. My suspicion is this fixes the detection but not all the functionality, so do report back on what is and is not confirmed to work.

PR #659, opened 2015-07-01
Fix fileignorecase and wildignorecase detection on OS X vim

mdippery:
Previously, this line caused an error on the version of vim that ships
with OS X (tested with OS X 10.10.4, vim 7.3). It appears that this
version of vim compiled and shipped with OS X supports neither
fileignorecase nor wildignorecase. Thus, this line was causing an error
when launching vim. The specific error was:

    E113: Unknown option: wildignorecase
    E15: Invalid expression: exists('+fileignorecase') ? &fileignorecase : &wildignorecase

This fix tweaks the conditional statement so that it runs properly if
neither option is present.

Signed-off-by: Michael Dippery <michael@monkey-robot.com>

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -558,7 +558,7 @@ function! s:buffer_commit() dict abort
 endfunction
 
 function! s:cpath(path) abort
-  if exists('+fileignorecase') ? &fileignorecase : &wildignorecase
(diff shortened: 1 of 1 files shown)


tpope:
Huh, looks like `'fileignorecase'` was added not long at all after `'wildignorecase'`. I'm going to drop support for the latter entirely.

PR #660, opened 2015-07-03
Ggrep suppport browsing and jumping to matching files

tomerf:
I created an interactive grep that opens a small window with the matches and allows to view the matching files and jump directly to the matches (file and line).

I based the code on Gblame and it looks pretty good IMO. I'm not an expert on vimscript (just enough to constantly tweak my own .vimrc) but as far as I tested it things are working well. Feel free to change it if you think it can be done better.

PS: I removed the Glgrep command. As far as I can see 'grep' is hard coded and (re)assigned to the 'cmd' argument without checking the original value, so there's really no difference. Same for Gllog, but I didn't touch it.
PS2: Add this line to .vimrc to grep the word under the cursor:
```nnoremap <leader>gg "zyiw:Ggrep \<<C-R>z\><CR>```

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1234,55 +1234,127 @@ function! s:Merge(cmd, bang, args) abort
   return exists('err') ? 'echoerr '.string(err) : ''
 endfunction
 
-" Section: Ggrep, Glog
+" Section: Ggrep
+
+augroup fugitive_grep
+  autocmd!
+  autocmd BufReadPost *.fugitivegrep setfiletype fugitivegrep
+  autocmd FileType fugitivegrep setlocal nomodeline | if exists('b:git_dir') | let &l:keywordprg = s:repo().keywordprg() | endif
+  autocmd Syntax fugitivegrep call s:GrepSyntax()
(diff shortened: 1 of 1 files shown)


tpope:
The quickfix is ugly but it works; I don't see a good enough reason for reinventing it.

Glgrep definitely works. Search for `a:cmd`.

tomerf:
What quickfix? The original grep is broken. I get a list of all matches and then it just go to the first one. No option to scroll and no option to choose what match I want, or even cancel and stay on the same file.

tomerf:
As for the lgrep, my bad. But llog doesn't work: ```let cmd = ['--no-pager', 'log', '--no-color']```

tpope:
:help :make
    :help :copen

tpope:
`a:cmd` and `cmd` are 2 unrelated variables.

tomerf:
Nice, the quickfix window is definitely better.

As I said, it doesn't open the window so I added an explicit call copen/lopen on ":Ggrep!" and made the grep call silent (seems to work only in gvim, not vim). Should I open a new pull request for that?

tpope:
FAQ addresses why `:Ggrep` works the way it does.

PR #661, opened 2015-07-10
ReplaceCmd: redirect stderr to tmp file / buffer

blueyed:
The option `status.showUntrackedFiles=all` used with `git status` for
`:Gstatus` might cause an error, which then causes fugitive to display an
empty status window / index file.

Redirecting the stderr output is useful in this case.

The generated command was:

    git --git-dir=/home/user/.dotfiles/.git -c 'status.displayCommentPrefix=true' -c 'color.status=false' -c 'status.short=false' -c 'status.showUntrackedFiles=all' status

The error from git is related to submodules being moved to another
subdirectory, where the relative "gitdir" now does not exist anymore:

    fatal: Not a git repository: vim/bundle.old.nobackup/CLEAN/colorscheme-base16/../../../.git/modules/vim/bundle/colorscheme-base16

While that's a Git / user error after all, fugitive should be more
helpful in that case by displaying the error.

This does not add the redirection for Windows (I am not sure if it would
work).

An alternative might be to redirect stderr to a separate file and then
display it using "echohl WarningMsg …".

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -2391,11 +2391,15 @@ function! s:ReplaceCmd(cmd,...) abort
         let prefix = 'env GIT_INDEX_FILE='.s:shellesc(a:1).' '
       endif
     endif
+    let redir = ' > '.tmp
+    if &shellpipe =~ '2>&1'
+      let redir .= ' 2>&1'
+    endif
     if s:winshell()
       let cmd_escape_char = &shellxquote == '(' ?  '^' : '^^^'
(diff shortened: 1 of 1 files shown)


tpope:
`2>&1` will work on Windows but it *won't* work in csh, and perhaps other circumstances. Probably best to use `'shellpipe'` as a heuristic.

blueyed:
Ok, amended.

I've seen now that there are several places handling this differently already - e.g. there's a check for 'csh' somewhere.

blueyed:
@tpope 
That should be good now?

PR #671, opened 2015-07-31
Detect 'git worktree add' checkouts

### LINKED ISSUE #655: Fugitive doesnt seem to understand git workdirs created by 'git checkout --to'
This command is reasonably new, git v2.4 or v2.4.1.

mathstuf:
The worktree feature of 2.5.0 stores a slimmed down .git directory under
the real .git directory:

    .git/worktree/$name

Under this directory, the files we care about:

    commondir -> relative path to the common .git directory
    gitdir -> absolute path to the .git file in the new worktree
    gitfile -> relative path to the .git file from the main worktree

Fixes #655.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -121,9 +121,26 @@ let s:abstract_prototype = {}
 
 " Section: Initialization
 
+function! fugitive#is_git_dir_reg(path) abort
+  " regular gitdir
+  return isdirectory(a:path.'objects') && isdirectory(a:path.'refs')
+endfunction
+
+function! fugitive#is_git_dir_worktree(path) abort
+  return getftype(a:path.'commondir') ==# 'file'
+endfunction
+
+function! fugitive#find_worktree_host(path) abort
+  return simplify(a:path.get(readfile(a:path.'commondir', '', 1), 0, ''))
+endfunction
+
 function! fugitive#is_git_dir(path) abort
   let path = s:sub(a:path, '[\/]$', '') . '/'
-  return isdirectory(path.'objects') && isdirectory(path.'refs') && getfsize(path.'HEAD') > 10
+  " The directory must have a HEAD file
+  if getfsize(path.'HEAD') <= 10
+    return 0
+  endif
+  return fugitive#is_git_dir_reg(path) || fugitive#is_git_dir_worktree(path)
 endfunction
 
 function! fugitive#extract_git_dir(path) abort
@@ -241,7 +258,10 @@ function! s:repo(...) abort
     if has_key(s:repos, dir)
       let repo = get(s:repos, dir)
     else
-      let repo = {'git_dir': dir}
+      let repo = {'git_dir': dir, 'git_commondir': dir}
+      if fugitive#is_git_dir_worktree(dir)
+        repo.git_commondir = fugitive#find_worktree_host(dir)
+      endif
       let s:repos[dir] = repo
     endif
     return extend(extend(repo, s:repo_prototype, 'keep'), s:abstract_prototype, 'keep')
@@ -255,6 +275,10 @@ endfunction
 
 function! s:repo_dir(...) dict abort
   return join([self.git_dir]+a:000,'/')
+endfunction
+
+function! s:repo_commondir(...) dict abort
+  return join([self.git_commondir]+a:000,'/')
 endfunction
 
 function! s:configured_tree(git_dir) abort
@@ -280,6 +304,9 @@ endfunction
 function! s:repo_tree(...) dict abort
   if self.dir() =~# '/\.git$'
     let dir = self.dir()[0:-6]
+  elseif filereadable(self.dir().'/gitdir')
+    let line = get(readfile(self.dir().'/gitdir', '', 1), 0, '')
+    let dir = line[:-6]
   else
     let dir = s:configured_tree(self.git_dir)
   endif
@@ -292,6 +319,8 @@ endfunction
 
 function! s:repo_bare() dict abort
   if self.dir() =~# '/\.git$'
+    return 0
+  elseif self.dir() =~# '/\.git/worktrees/[^/]*$'
     return 0
   else
     return s:configured_tree(self.git_dir) ==# ''
@@ -322,16 +351,19 @@ function! s:repo_translate(spec) dict abort
     return 'fugitive://'.self.dir().'//0/'.a:spec[1:-1]
   elseif a:spec =~# 'HEAD\|^refs/' && a:spec !~ ':' && filereadable(self.dir(a:spec))
     return self.dir(a:spec)
-  elseif filereadable(self.dir('refs/'.a:spec))
-    return self.dir('refs/'.a:spec)
-  elseif filereadable(self.dir('refs/tags/'.a:spec))
-    return self.dir('refs/tags/'.a:spec)
-  elseif filereadable(self.dir('refs/heads/'.a:spec))
-    return self.dir('refs/heads/'.a:spec)
-  elseif filereadable(self.dir('refs/remotes/'.a:spec))
-    return self.dir('refs/remotes/'.a:spec)
-  elseif filereadable(self.dir('refs/remotes/'.a:spec.'/HEAD'))
-    return self.dir('refs/remotes/'.a:spec,'/HEAD')
+  elseif a:spec =~# 'HEAD\|^refs/' && a:spec !~ ':' && filereadable(self.commondir(a:spec))
+    return self.commondir(a:spec)
+  endif
+  if filereadable(self.commondir().'/refs/'.a:spec)
+    return self.commondir().'/refs/'.a:spec
+  elseif filereadable(self.commondir().'/refs/tags/'.a:spec)
+    return self.commondir().'/refs/tags/'.a:spec
+  elseif filereadable(self.commondir().'/refs/heads/'.a:spec)
+    return self.commondir().'/refs/heads/'.a:spec
+  elseif filereadable(self.commondir().'/refs/remotes/'.a:spec)
+    return self.commondir().'/refs/remotes/'.a:spec
+  elseif filereadable(self.commondir().'/refs/remotes/'.a:spec.'/HEAD')
+    return self.commondir().'/refs/remotes/'.a:spec.'/HEAD'
   else
     try
       let ref = self.rev_parse(matchstr(a:spec,'[^:]*'))
@@ -359,7 +391,7 @@ function! s:repo_head(...) dict abort
     return branch
 endfunction
 
-call s:add_methods('repo',['dir','tree','bare','translate','head'])
+call s:add_methods('repo',['dir','commondir','tree','bare','translate','head'])
 
 function! s:repo_git_command(...) dict abort
   let git = g:fugitive_git_executable . ' --git-dir='.s:shellesc(self.git_dir)


mathstuf:
So this at least gets `Gstatus` working. There's some work that needs to be done for other commands (namely `Gdiff` dumping out object contents rather than what is important).

mathstuf:
OK, so I've got `Gdiff` to work now. Please test heavily :) . I'll try poking about with it over the next few days to see if anything weird pops up.

tpope:
I don't currently use worktrees, so if anyone else could test this feature and report back that would be great.

glittershark:
I don't use them but holy crap do I *want* to. They look great. I'll check this out!

glittershark:
This isn't working for me at all - none of the `:G...` commands are defined in files inside the worktree

mathstuf:
Hrm. Means it isn't detecting that it's a git tree. What happens inside of `fugitive#is_git_dir` that makes it fail? (`:debug fugitive#is_git_dir(".")` and then step once or twice and run the subexpressions)

mathstuf:
Err, `echo ...` the subexpressions.

plugin/fugitive.vim
```diff
@@ -123,7 +123,7 @@ let s:abstract_prototype = {}
 
 function! fugitive#is_git_dir(path) abort
   let path = s:sub(a:path, '[\/]$', '') . '/'
-  return isdirectory(path.'objects') && isdirectory(path.'refs') && getfsize(path.'HEAD') > 10
+  return ((isdirectory(path.'objects') && isdirectory(path.'refs')) || getftype(path.'gitdir') ==# 'file') && getfsize(path.'HEAD') > 10
```
jwhitley: What does this change mean?  `path`, i.e. a potential `.git` repo directory, is never going to contain a normal file called `gitdir`.  A secondary worktree contains a normal file `.git` at its root that contains text of the form:

 `gitdir: /path/to/repo`

But that use case already appears to be handled by `fugitive#extract_git_dir()`, below at [line 161](https://github.com/tpope/vim-fugitive/pull/671/files#diff-ce13abb0779d494c65de144c88ac3992R161).  (Though I note that, despite that code, vim-fugitive doesn't correctly recognize a secondary worktree as having a valid git directory repo path.)

mathstuf: This function gets the `/path/to/repo` as its argument, not the path to the `.git` file in a worktree. A worktree does neither contains an `objects` nor `refs` directory, so the `/path/to/repo/.git/worktrees/foo` it gets fails the checks here and fugitive says that the directory is not run by `git`.

jwhitley: Thanks, understood. Perhaps it's worth breaking that logic out structurally,  e.g. an `if` case that handles a usual `.git` directory and an `else` for worktrees, and comments for each.  Or at least throwing down some comments that illuminate things a bit.  Because this increasingly complex and undoc'ed compound logic test is inscrutable, especially with the addition of a new kind of input (worktree state dir) from what's been received previously (only a .git dir).

jwhitley:
I've done preliminary testing, hitting `:Gstatus` in the following worktree types:

 * git repo without associated worktrees
 * git repo with external worktree defined by $GIT_DIR environment variable (aka "fake bare repo" as used by [vcsh](https://github.com/RichiH/vcsh) and some other tools)
 * git repo, parent worktree with additional worktrees
 * an additional worktree dir of the above

In all cases, things worked as expected.  E.g. In each case, I saw the correct branch and/or repo state (e.g. detached head), file status, etc.  I'll continue to work with @mathstuf's branch and report back if I find anything.

jwhitley:
@glittershark, alternatively, if you can provide repo steps, I'll pitch in and debug.  Something ala:

1.  `git clone https://url/to/foo`
2.  `cd foo`
3.  `git worktree add ../foo2`
4.  `cd ../foo2`
5.  `vim`
6.  In vim: `:Gstatus`

Also, can you confirm what `git version` reports for you?  Thanks!

mathstuf:
FWIW, I know there is a bug with `:Gcommit`, `:Gwrite`, and associated commands. They are using the wrong `index` file.

mathstuf:
I'm thinking of adding a second call to the repo object for getting different parts of the `.git` directory, one for read-only (always the main `.git`) and read-write (the worktree path or the `.git` dir). All places where the path is queried will need updated too.

mathstuf:
OK, some light testing indicates that `:Gwrite` and `:Gadd` are working now. `:Gcommit` still needs some work (it's using the common directory's index file).

mathstuf:
So I've been unable to figure out where `b:git_dir` is set in the `COMMIT_EDITMSG` buffer for `:Gcommit`. If anyone knows where this is, it needs to be updated to learn how to detect a worktree gitdir and set that instead.

jwhitley:
Some working notes as I dig in.

The places `b:git_dir` is assigned in the `git-workdir-support` branch are in:

1.  [`fugitive#detect()` at L201](https://github.com/mathstuf/vim-fugitive/blob/git-workdir-support/plugin/fugitive.vim#L201)
2.  [`s:BufReadIndex()` at L2460](https://github.com/mathstuf/vim-fugitive/blob/git-workdir-support/plugin/fugitive.vim#L2460)
3.  [`s:BufReadIndexFile()` at L2552`](https://github.com/mathstuf/vim-fugitive/blob/git-workdir-support/plugin/fugitive.vim#L2552)
4.  [`s:BufReadObject()` at L2606](https://github.com/mathstuf/vim-fugitive/blob/git-workdir-support/plugin/fugitive.vim#L2606)
5.  [`autocmd BufNewFile,BufReadPost *` at L2715](https://github.com/mathstuf/vim-fugitive/blob/git-workdir-support/plugin/fugitive.vim#L2715)

The BufRead* functions all set from `s:repo().dir()`.  IIRC, that's covered by this patch already.

The autocmd is set from an element of `s:temp_files`, the `dir` element of which is initialized in two places, by `buffer.repo().dir()` (L1394) and `s:repo().dir()` (L1995), so that also seems to be covered by changes to the repo class.

That leaves `fugitive#detect()`, which in turn depends on `fugitive#extract_git_dir()`.

jwhitley:
@mathstuf I just gave `:Gcommit` a try from an up-to-date version of your branch, and it seems like it's working?  A test workflow and repro steps below.  What am I missing?

screen shot 2015-08-09 at 9 18 27 pm

**Repro steps**
1.  `git init fugitive-test`
2.  Create `a.txt`, content "this is a".
3.  `git add a.txt`
4. `git commit -m 'added a'`
5.  Create `b.txt` content "this is b"
6.  `git add b.txt`
7.  `git worktree add ../ft2`
8.  `cd ../ft2`
9.  Create `c.txt` content "this is c"
10.  `git add c.txt`

In each of `fugitive-test` and `ft2`, run vim and invoke `:Gcommit`.  Per the screenshot, I see `b.txt` as a new file in the index in `fugitive-test` and `c.txt` as a new file in `ft2`, and the paths shown for the respective COMMIT_EDITMSG files are correct (main repo dir for `fugitive-test` and worktree dir for `ft2`).

mathstuf:
Did you try actually committing? It opens the main .git/index file in the raw and the commit does not happen. Inspecting, the b:git_dir variable is wrong in the COMMIT_EDITMSG buffer. Tracing with -V20 did not show where it was set (I aslo echoed after each let and it never printed anything (not even the wrong directory). I guess it is the autocmd one since echo might not work there?

jwhitley:
Errr, bad news ... it works for me, across different systems, and vim & neovim ... ??:

I did commit, in both the main and workdir repos, for that test.  That worked fine.  Just repro'ed on a different host.  Again, worked just fine in both the main and child workdirs.  I check after each commit with `git log`, and all appears as it should.  `b:git_dir` is likewise correct.

For good measure, I just re-tested successfully with both:

 * neovim, Commit neovim/neovim@5fc4c2d442f01ab53fa95c7732c834950d283190 (circa 2015-Aug-4)
 * vim 7.4.712

(all with git at v2.5.0, obviously.)

glittershark:
lol. My problem is that I mucked up my rtp so that my checkout of Fugitive was pointing at tpope/master not this branch. So, yeah, this LGTM :stuck_out_tongue_closed_eyes:

tpope:
Can someone give me the tl;dr explanation of the difference between `dir` and `dir_store`, with concrete examples? I have yet to use worktrees.

jwhitley:
@tpope 

 * `dir` is the path git wants as `$GIT_DIR` for the current worktree.
 * `dir_store` must always be the path to the main repository root, where things like `refs/`  and `objects/` are located.

In a "classic" git repository, these two are always the same.  But when `$cwd` is inside an added worktree, they're now different.  Using my example above, in `/path/to/ft2` these would be:

 * `repo.dir()` : `/path/to/fugitive-test/.git/worktrees/ft2`
 * `repo.dir_store()` : `/path/to/fugitive-test/.git`

All of this is transparent to a user of git porcelain or plumbing commands, but not to fugitive since it's aware of the repository layout.

Refs: see [gitrepository-layout(5)](http://www.git-scm.com/docs/gitrepository-layout), specifically the entries starting with `worktrees`, toward the end.

mathstuf:
Actually, this should fix any `.git` dir with `commondir` within it (it is not worktree-specific). I'd like to know why it is broken for me, but if it's just me, I'm fine with making another pull request to fix whatever I'm hitting after this is merged :) .

@jwhitley What git version and distribution (I'm using Rawhide)?

jwhitley:
@mathstuf Good point about `commondir`, thanks. @tpope, from my review, I'll agree that the new code should properly respect the presence of commondir above and beyond its use in worktrees.

I'm using git 2.5.0 (since I'm using worktrees) installed via homebrew on OS X (10.10.4).

Just for kicks, I repro'ed successfully again with an Ubuntu 14.04 VM, git 2.5.0, vim 7.4.52.  In all cases, I'm using my standard homedir+vimrc setup.  AFAICT, there are no interesting or unusual vim-fugitive related settings, just the usual spate of <leader> bindings for its commands.

So yeah, I dunno what's going on on your end, but it's just plain working for me.  Nice work!  ;-)

jwhitley:
A clarification based on something that just clicked for me: `dir_store` is not only pulled from the `commondir` file, but it's semantically identical.

So for clarity, I'll propose that `dir_store` just be renamed `commondir` everywhere it appears in the code.  Then the code will be aligned with git's terminology for the same concept.  Also, it's then straightforward to add pointers to `$GIT_COMMON_DIR` in [git(1)](http://git-scm.com/docs/git) and `commondir` in [gitrepository-layout(5)](http://git-scm.com/docs/gitrepository-layout) which will help illuminate what this part of the code is all about.

mathstuf:
Renamed `dir_store` to `commondir`. I tried `git` master to no avail too :( .

jwhitley:
Good grief.  Tomorrow I'll have a go at creating a pristine VM with a stripped down .vimrc to see if I can get it to fail.

I suppose it's somehow possible that something in the pseudointelligent virus that my [.vimrc setup](https://github.com/jwhitley/vimrc) has evolved into might be unborking things?  :grimacing:

mathstuf:
Well, it apparently works for @glittershark too, so I'm tempted to say it's on my end. To be honest, my vimrc setup is probably just as complicated, of not more :) . As you toss plugins on the heap, things always get hairy. I may have some time at work to test with the worktrees I've been using there, but no guarantees.

tpope:
Let's make those 3 new functions private.

jwhitley:
AH-HA!  I can reproduce the `:Gcommit` issue.  At least, `:Gcommit` doesn't commit, and there's an `echom` that just says `fugitive:`.

I made a repo with Vagrantfile which produces an Ubuntu 14.04 VM with git 2.5.0+, and a minimal vimrc setup, just vim-plug and this branch of vim-fugitive.  See the README at [jwhitley/fugitive-worktrees](https://github.com/jwhitley/fugitive-worktrees) for setup and usage instructions.  It's trivial to tweak, so shout out here if there's anything I can do that'd help out.

The actual `~/.vim/` and `~/.vimrc` are symlinked into place from the Vagrant shared folder.  I.e. you can edit the versions in this repo on the host system, and the VM will pick up your changes immediately.  This will be useful for hacking on the plugin (in `fugitive-worktrees/.vim/plugged/vim-fugitive`) with a full vimrc, while testing it out and debugging in the VM.

So yeah, there *is* something mysterious in my environment that makes this bug not appear.  Bizarre!

I'll have more time later to dig in, but for now this provides a nice, repeatable environment to hunt this bug.

mathstuf:
Cool! I noticed a typo in the vimrc, so maybe that's from yours as well? Can't imagine that encoding saves you from this bug though…

jwhitley:
Sorry, I'm completely not seeing the typo.  Can you be more specific?  Both the `set encoding=utf-8` and `scriptencoding utf-8` lines seem OK, if that's what you're talking about... ?  (And yeah, the top three lines of boilerplate are inherited from my setup.)

mathstuf:
[Here](https://github.com/jwhitley/fugitive-worktrees/commit/21282e05402c0de366e271158622a07af0b31ed6#diff-29697a64bb002da9d51f19e96d0166dcR3).

jwhitley:
Oh heck, that's vestigial and not used.  I accidentally committed that during early indecision about how to setup the repo.  The actual ones used are in the repo's root directory.

mathstuf:
So something changed in git and/or vim and this now works for me.

jwhitley:
@tpope @mathstuf I just tried again with my VM, and I'm still getting an issue with committing in a worktree directory in my test VM as listed above.  This test was against git 2.6.0, suggesting that the problem isn't something related to the underlying version of git, which is also what I'd expect from fugitive's code and this PR.

Given that this problem only affects some system configurations, but works well and completely on many others, I suggest that this PR be merged and a new issue be spun off to track and fix the failing cases, based on my comment re: the repro VM above.

mathstuf:
So one other thing I've seen is that git is rewriting `gitdir` to be…wrong in some cases. I need to be more diligent about noticing when that happens for a full report. Thinking about it a bit, it might be related to trying to `git checkout $worktree_branch` which then fails, but may have changed `gitdir`? *That* is an upstream git bug though.

jwhitley:
@mathstuf Interesting, I'll keep an eye out for it.  But to be clear, that's unrelated to the cases where this PR has issues.  `gitdir` is easily verified as correct in my test setup.

mathstuf:
Ah, that bug was already fixed upstream in v2.5.2: git/git@82fde87ff3be8d10854df18964e5816417a1d7cf

sophacles:
Hi all - Is there anything I can contribute to move this PR/issue along? I use a lot of worktrees and would love fugitive support...

mathstuf:
I'll remove the `WIP` tag. I haven't had problems with it, but there is some lingering error that hasn't been tracked down.

jwhitley:
:+1:  I'd like to see this merged as well.

grodzik:
I've switched to your branch and only thing I get when running Fugitive command (:Gstatus for example) is 'fugitive: no work tree'. Works as expected in normal repo. My worktrees are located under '<git-root>/worktrees/', if that changes anything.
Any idea what could be wrong, is it possible that one of other plugins is messing fugitive's workflow?

grodzik:
Sorry for the alarm, after some more testing it's working ok. Problem was with gitdir in .git/worktrees/<worktree>/gitdir. I'm pretty sure that I had git 2.6.2 at the time of creating that work tree and now I have git 2.6.3, which makes me think that [337e51ce](https://github.com/git/git/commit/337e51cedfdf1f98e1c151b72f06c7ca8bcea0fd) or [31e26ebc](https://github.com/git/git/commit/31e26ebcb512ad234d9725120fb4f5670ff75625) fixed the issue with it.

knirch:
@grodzik or one could do something silly like me and move the worktree, which caused the same error message. Regardless, thank you for your comment, that allowed me to find why my worktree wasn't working.

lilydjwg:
I get 'fugitive: no work tree' when the gitdir file contains a relative path (`.git`). It works after I manually change it to the absolute path.

I don't know why it is relative. I've tried to create other worktrees and they are absolute. I've checked my system logs but I haven't upgraded git. It's strange. I'm on Arch Linux, git 2.6.4.

jwhitley:
@lilydjwg I'm not sure how you got a relative path, but that's a clear violation of the [gitrepository-layout(5)](http://git-scm.com/docs/gitrepository-layout) manpage specs for `worktrees/<id>/gitdir`:

> A text file containing the absolute path back to the .git file that points to here.

IIRC, Fugitive, this branch included, never directly manipulates repository or worktree files; it just reads them.  So I expect something else, e.g. a bug in git or some other tool, is causing this.

knirch:
They're tracking a gitdir mangling bug; http://thread.gmane.org/gmane.comp.version-control.git/283998

lilydjwg:
@knirch thanks, that's the bug happening to me.

basepi:
Hitting this myself. My most-used command is :Gblame and it doesn't work in a worktree. Hopefully we can get this finished up and merged soon.

mathstuf:
Not sure that it should block this branch; it's a git bug that can't be worked around (there's not enough information to make the full path again).

zhimsel:
So I don't know if this helps, but it seems that `vim-fugitive` isn't even loading its commands properly if launched from a worktree... I get `E492: Not an editor command: Gstatus` when trying to do a `Gstatus`. As far as I can tell, all the other `G...` commands aren't there either. However, `vim-plug` reports that the `vim-fugitive` is loaded, and running `vim` with a debug log seems to agree (there are no errors that I can find). The commands just... aren't there.

They work fine when run in a non-worktree repo, though.

```bash
$ git --version
git version 2.7.0

$ vim --version
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Jan 31 2016 16:19:26)
MacOS X (unix) version
Included patches: 1-1190
Compiled by Homebrew
```

mathstuf:
If you're not using this branch, fugitive won't detect that it's a git repository and will not add its bindings.

zhimsel:
Ah, okay, that makes sense (duh). Thanks.

sergei-dyshel:
Confirming: I'm on Git 2.7.1 and `git worktree add ../....` created a relative path in `gitdir` file.

therealprof:
Just wanted to chime in. The lack of worktree support is driving me nuts as I'm using that feature extensively, so I went ahead and cherry-picked the change 51c4f66aeac3f4a5df08b5be5e760c2b3bc87a33 and I'm happy to confirm that this works nicely for me so please consider applying it.

Regarding the git worktree problem which was mixed in here: Indeed git tried to be "smart" by trying to detect manually moved directories which did not work and caused havoc so if you do something as simple as a git reset --hard it work rewrite the path to the current directory. This automagic has been removed in the just released 2.7.2 and now works just fine; however if the relation is already messed up you'll have to fix up all the worktrees manually.

balaclark:
I've also tested this locally and can confirm that this commit fixes `:Gblame`

tpope:
What precisely is the difference between `dir` and `commondir`? Is `commondir` standard Git terminology or something invented?

therealprof:
I think the commit sums it up pretty well:

> Under this directory, the files we care about:
> 
>     commondir -> relative path to the common .git directory
>     gitdir -> absolute path to the .git file in the new worktree
>     gitfile -> relative path to the .git file from the main worktree

Those are standard git files.

tpope:
Okay the pull request description answers my first question. But if `commondir` is relative then why is it never made absolute? Does everything break if the current working directory changes?

therealprof:
The worktrees contain only a file named .git rather than the usual directory .git only pointing to the worktree directory inside the main repository clone. commondir and gitdir only exists inside the .git directory in the main repository so using a relative link (as chosen by the git team) should be safe. The reason for it's existence is actually not clear to me as the directory structure seems pretty fixed but maybe there're plans to allow a tree of worktrees rather than just a flat list in the future.

mathstuf:
The worktree's actual checkout's `.git` should be absolute to the worktree `.git` directory contents. This `.git` directory is special in that is is meant to always be under a full `.git` directory under `worktrees/$name`. I don't think `commondir` is going to be anything other than `../..`, but who knows. I suppose `$name` could have directory separators in it…

tpope:
If `commondir()` is relative, then this has a broken assumption about the current working directory, full stop:

    if filereadable(self.commondir().'/refs/'.a:spec)

mathstuf:
Hmm, seems I have a minor diff around there; I'll have to see about syncing up again.

In any case, it isn't a real symlink, but is a file with content of a relative path (because Windows probably).

therealprof:
Yes, indeed. It's not a link but a text file containing part of the path. As far as I can tell find_worktree_host () reads the file and creates an absolute path keyed of the gitfile so your 

> if filereadable(self.commondir().'/refs/'.a:spec)

does exactly the right thing.

tpope:
Okay I see what's going on. Try the `worktree` branch, which eliminates some unnecessary IO (and seems a lot simpler, to boot).

therealprof:
Doesn't look simpler to me but works just as fine. ;)

tpope:
Okay, well I have no idea if it's a 100% worktree solution, but I feel pretty confident it doesn't introduce any regressions, so I'm merging.

basepi:
Looks like you closed, rather than merging?

tpope:
I merged my own branch not this one.

basepi:
Cool, sorry, missed that.

sophacles:
So this may have been premature - on osx, with:
* git 2.7.2
* macvim 7.4 (run as cli vim)

both installed via `homebrew`

`:Gstatus` is showing all the files in the repo as "deleted", none as modified, but only in a worktree. Regular branches work fine.

therealprof:
I don't have that problem. Also on OS X with vim 7.4.1401 and git 2.7.2 installed via homebrew. There's absolutely no difference between git status on the commandline and :Gstatus .

sophacles:
Looking at master, vs this PR, it looks like what was committed as "support worktrees" was different than the file in here as well.... I'm using current master. Perhaps a new ticket?

tpope:
Same setup works fine for me. Open a new issue I guess.

davidmaxwaterman:
I'm seeing something that looks very similar to this with git submodules. Could it be the same thing? The submodule directory contains a .git file with a gitdir: line that is a relative path to "the current directory" inside the parent repository's .git directory.
I'm using 7.4 1-712 on ubuntu 15:10 with git 2.5.0.

sophacles:
The issue I reported above turned out to be weird relative path issues that were artifacts of bugs in older git versions. I had to manually fix these to be absolute paths in long-running worktrees. New worktrees are just fine out of the box. Is there a way for you to update to git 2.7 @davidmaxwaterman ?

davidmaxwaterman:
Yes, I used the ubuntu ppa to update to 2.7.3. It's not made any difference to Gblame. I still get the error :

```
fugitive: fatal: no such path 'parent_git_root/..../git_root/path_to/current_file' in HEAD
```

Gdiff behaves as if the file didn't exist before the current version (ie left buffer is blank), and the buffer name is :

```
fugitive:///path_from_fs_root_to_parent_git_root/.git//0/parent_git_root/..../git_root/path_to/current_file
```

I hope this makes some sense. I guess it should be easy enough to reproduce just by creating a couple of git repositories, cloning one, and then making a submodule of the other one. I might give this a try when I get (another) spare moment.

davidmaxwaterman:
Any further thoughts on the similarity of this issue when using git submodules?

jwhitley:
@davidmaxwaterman I'd just open a new issue.  At this point, it's hard to speculate what's going on with submodules in your situation.  Repro steps (and maybe a repro repository, if possible) would help hugely.

davidmaxwaterman:
OK, I did as you requested :
https://github.com/tpope/vim-fugitive/issues/769
Thanks.

PR #683, opened 2015-08-27
Glob: stash references and fix special chars

blueyed:
This adds globbing for `stash`, `stash@{0}` etc, and fixes the escaping of filesnames - especially for files that `git-ls-files`/`git-ls-trees` considers to be special (by using `-z`).

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -404,6 +404,11 @@ function! s:repo_superglob(base) dict abort
     if a:base !~# '^/'
       let heads = ["HEAD","ORIG_HEAD","FETCH_HEAD","MERGE_HEAD"]
       let heads += sort(split(s:repo().git_chomp("rev-parse","--symbolic","--branches","--tags","--remotes"),"\n"))
+      " Add any stashes.
+      if filereadable(s:repo().dir('refs/stash'))
+        let heads += ["stash"]
+        let heads += sort(split(s:repo().git_chomp("stash","list","--pretty=format:%gd"),"\n"))
+      endif
       call filter(heads,'v:val[ 0 : strlen(a:base)-1 ] ==# a:base')
       let results += heads
     endif
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -424,7 +424,7 @@ function! s:repo_superglob(base) dict abort
 
   elseif a:base =~# '^:'
     let entries = split(self.git_chomp('ls-files','-z','--stage'),"")
-    call map(entries,'s:sub(v:val,".*(\\d)\\t(.*)",":\\1:\\2")')
+    call map(entries,'s:sub(v:val,".*(\\d)\\t(.*)","\\=\":\".submatch(1).\":\".fnameescape(submatch(2))")')
     if a:base !~# '^:[0-3]\%(:\|$\)'
       call filter(entries,'v:val[1] == "0"')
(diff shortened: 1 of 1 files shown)


blueyed:
@tpope 
I'd appreciate feedback on this.. :)

tpope:
Not sure what your issue is with the literal `<C-A>` but that needs to be fixed.

blueyed:
> issue is with the literal `<C-A>`

What do you mean?

tpope:
Why the hell is there a literal binary character in your diff?

blueyed:
Where?  I cannot see it in https://github.com/tpope/vim-fugitive/pull/683/files - can you post a link to the line - or the line number, please?

tpope:
No

blueyed:
I've seen it in https://github.com/tpope/vim-fugitive/pull/683.patch.

Should be fixed now.
(was from https://github.com/blueyed/vim-fugitive/commit/ec3cd544a46012074559d084c48bf93ce0d1c5d8)

tpope:
Your last commit has a long, obsolete explanation of why you used `<C-A>`.

tpope:
And the other commit is completely broken.

blueyed:
> Your last commit has a long, obsolete explanation of why you used <C-A>.

It does not mention <C-a>, but I've used it for `NUL`.  I've found no other way to handle/use NUL inside of Vim.  What do you recommend?

> And the other commit is completely broken.

Do you mean https://github.com/blueyed/vim-fugitive/commit/e8335ebe594cc94cebdd601956de832975190b4f?

tpope:
You say `It would be better to not include the NUL as <C-V><C-A>` in a commit that contains neither, best I can tell.

Yes.

blueyed:
I see, so I probably mixed up the "comes from" comment or the commit message.

But given that `<C-a>` is required: is there a better way to handle `NUL`?

What do you mean with "completely broken" wrt to the last commit?

tpope:
I mean I do `:Gedit master:<Tab>` and it's full of garbage.

blueyed:
> I mean I do :Gedit master:<Tab> and it's full of garbage.

That's likely because of the (now missing) `<c-a>`.

This PR uses NUL-bytes for splitting, but I've removed the proper handling of it - because I've thought the `<c-a>` was there by accident.

Throws us back to the question if there's a better way to handle them in Vim.

tpope:
Open a fresh request once you get your shit together.

blueyed:
So that means there's no better way than a literal `<c-a>` to handle `NUL` in Vim, and the reason to close the PR is that I've added the `long, "obsolete" explanation` for it to the wrong commit?!

I'd fix the PR by re-adding the `<c-a>` (I've asked too many times already if there would be a better way without an answer), and edit the commit messages.  Which boils down to the same code, with different commit messages: #695.

tpope:
There are half a dozen better ways, clearly documented with the syntax for strings.

blueyed:
I'd really appreciate a pointer then.

`nr2char(0)` for example does not work, and so doesn't `nr2char(10)`. (But I've only tested it in Neovim FWIW.)

tpope:
`"\001"`

blueyed:
Thanks, that works.
But I fail hard at finding it in the help, looking at `:h expr-strings` would still leave the impression that it should be `\000`, apart from that this is meant to end a string.

PR #689, opened 2015-09-20
Add support for git core.commentChar configuration

### LINKED ISSUE #547: Add support for core.commentchar configuration
Git 1.8.2 allows setting a configuration called `core.commentchar` ([see announcement](http://article.gmane.org/gmane.linux.kernel/1443021)). This allows users to change the default `#` to something else, so that commit subjects may e.g. start with issue references.

It would be great if fugitive could change the syntax highlighting of commit messages accordingly.

aschrab:
Detect the comment character being used by git by looking at the first character included in the status buffer.  Use that character rather than a hard-coded `#` when operating in `:Gstatus` buffers.

I don't think this is actually ready to merge yet. It fails rather spectacularly if the comment character is set to `\`,  `^`, or `]`. But, as long as I don't pick one of those characters it is working well for me, along with the changes from tpope/vim-git#31. If I can get these issues fixed, I believe this would resolve #547.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -778,7 +778,8 @@ function! fugitive#reload_status() abort
 endfunction
 
 function! s:stage_info(lnum) abort
-  let filename = matchstr(getline(a:lnum),'^#\t\zs.\{-\}\ze\%( ([^()[:digit:]]\+)\)\=$')
+  let comment = b:comment_char
+  let filename = matchstr(getline(a:lnum),'^['.comment.']\t\zs.\{-\}\ze\%( ([^()[:digit:]]\+)\)\=$')
   let lnum = a:lnum
   if has('multi_byte_encoding')
     let colon = '\%(:\|\%uff1a\)'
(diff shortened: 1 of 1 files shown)


charlie-wasp:
What about this PR? I am using backtick for my git commentchar, so this improvement will be very helpful

aschrab:
@charlie-wasp As noted in the original description, I don't think this is ready for merging as it is, since it is known to behave very badly with some configurations. And before the updates that I just pushed it also didn't work for some actions (such as hitting `enter` on a file name in the status window).

I think this now solves the issue with most comment character configurations, but I think the problems with some choices of comment character still need to be solved before I'd consider this ready for merging.

aschrab:
@tpope With the latest commit that I've added here, all of the issues I've seen with this seem to be fixed. I've been using the original change from here for several months with `!` configured as my comment character, without any major problems. The only problem I'd experienced occasionally was that the `enter` binding wasn't working, but that should be fixed with commit c82eacd.

PR #695, opened 2015-10-05
Better handling of stashes

blueyed:
This adds globbing for `stash`, `stash@{0}` etc, and fixes the escaping of filesnames - especially for files that `git-ls-files`/`git-ls-trees` considers to be special (by using `-z`).

Follow-up (with adjusted commit messages) of #683.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -461,8 +461,8 @@ function! s:repo_superglob(base) dict abort
     return results
 
   elseif a:base =~# '^:'
-    let entries = split(self.git_chomp('ls-files','--stage'),"\n")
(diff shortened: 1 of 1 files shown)


blueyed:
@tpope 
Any input on this one?
It is required for correct completion of stashes.

tpope:
Needs to be redone, but I'm ready to reconsider.

blueyed:
This is mostly about (not) escaping and using NUL to split output.
It would need to pass the char to be split by to `s:LinesError` likely.

However, handling of stashes appears to be good/better on master by now.

tpope:
There's `s:NullError()` for that.

PR #702, opened 2015-10-16
Fix diff sometimes opening in preview window

wilywampa:
If the previous window no longer exists when Gedit is called, the attempt to change windows with 'wincmd p' fails and 'wincmd w' should be used instead. For example, if the preview window is 1, middle window is 2, and bottom window is 3  and closing window 3 makes window 1 active, `Gedit` will not switch to window 2 like it should.


tpope:
Thanks

PR #703, opened 2015-10-19
Encode URLs before passing them to netrw

keith:
This fixes a bug where characters such as # were escaped by netrw which 404s on GitHub

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -2273,6 +2273,7 @@ function! s:Browse(bang,line1,count,...) abort
       call s:throw("Instaweb failed to start and '".remote."' is not a supported remote")
     endif
 
+    let url = s:url_encode(url)
     if a:bang
       if has('clipboard')
         let @* = url
@@ -2387,6 +2388,10 @@ function! s:instaweb_url(opts) abort
     let url .= '#l' . a:opts.line1
   endif
   return url
+endfunction
+
+function! s:url_encode(str)
+  return substitute(a:str,'[^A-Za-z0-9_.:/#~-]','\="%".printf("%02X",char2nr(submatch(0)))','g')
 endfunction
 
 if !exists('g:fugitive_browse_handlers')


tpope:
I think we need to escape the path portion of the URL only, which means `s:github_url` and `s:instaweb_url` both need to call it. Otherwise `?` parameters and the `#` fragment pose problems, not to mention `user:pass@host` constructions 3rd party providers might use.

keith:
I've updated this to escape just the path portion in the relevant places. Let me know if I missed anything!

tpope:
So one thing that complicates this is third party support. Ideally, we'd expose the URL encode function, but I think for now I'd rather go the lazy route and just do some conservative escaping of the URL as a whole.

keith:
Am I missing a place for the other URLs that need escaping?

tpope:
At least 3 that I know of:

* https://github.com/shumphrey/fugitive-gitlab.vim
* https://github.com/tommcdo/vim-fubitive
* https://github.com/tpope/vim-rhubarb

Like I said, let's do minimal escaping of the final URL instead, for now.

keith:
Sorry I meant what entry point should we escape at? Possible https://github.com/keith/vim-fugitive/blob/14b854282b14892a55e070c9d5cfae4e9706eab3/plugin/fugitive.vim#L2258 ?

tpope:
Above the subsequent `if a:bang` I would say.

keith:
Added!

tpope:
Let's do that spot *only* and avoid double escaping by only hitting obvious problems like spaces.

keith:
Oh nice call. Sorry this is taking so much input. Updated!

tpope:
Yeah this still misses the mark. I *don't* want to escape e.g. `%` because then if a plugin does the right thing we'll end up escaping twice. 

I'm pushing up a version that escapes spaces and angle brackets only. Follow up if there are other characters I should include. I can't find the authoritative list in netrw.vim.

keith:
Awesome, thanks for spending time on this! I'll let you know if I find any other chars this doesn't cover.

PR #706, opened 2015-10-22
Add FAQ entry for adding custom github domains.

codemariner:
This might save some folks time from hunting down an answer.  It would have for me! :-)


tpope:
This is a Rhubarb concern now (where I believe it's sufficiently documented but feel free to send a patch if not).

PR #714, opened 2015-11-06
Document StageUndo key map (U) in :Gstatus

thallada:
Not sure if there was a reason this was left undocumented. I didn't know about this feature until I searched the Github issues. I think it is essential for a lot of people's workflows and should be documented.

### DIFF
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -57,6 +57,9 @@ that are part of Git repositories).
                         q     close status
                         r     reload status
                         S     |:Gvsplit|
+                        U     |:Git| checkout
+                        U     |:Git| checkout HEAD (staged files)
+                        U     |:Git| clean (untracked files)
 
                                                 *fugitive-:Gcommit*
 :Gcommit [args]         A wrapper around git-commit.  If there is nothing


camilopayan:
:+1:

rgrinberg:
Did not know this shortcut. +1!

tpope:
I initially left this undocumented because I hate the mnemonic and was hoping to change it, but at this point I don't think that's going to happen.

Untracked files are deleted not cleaned, but seeing as how `git clean` takes a path, it sounds like a much better choice. Anybody have thoughts on what flags it should take?  Would like to take care of this first so the documentation is accurate at time of merge.

firedev:
I am using it all the time, the only problem is when you are rebasing and the file is deleted in the base branch and you want to delete it when resolving the conflict you have to first stage and unstage it and only then you can `U` get rid of it.

You also better remeber the name of it when you are pressing `-` the first time to stage it as you need to search for it in the list to unstage.

tpope:
So should we be calling `git rm` on unmerged files?

firedev:
I'd say so, in my case that would really help.

tpope:
I went

firedev:
Wow, that was quick, thanks!

justinmk:
Glad to this. `U` is very useful. Thanks.

PR #715, opened 2015-11-07
Add `:Gtdiff` command(s) to open diff in a new tab

wadkar:
As discussed in PR #413 and #319 this feature adds relevant command and
documentation to open Git diffs in new tab.

Note that like #413, the new tab is created before the current working tab.
After playing around for a while, it appears that closing the new diff tab
should take you back to the original working tab.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1687,6 +1687,9 @@ endfunction
 call s:command("-bang -bar -nargs=* -complete=customlist,s:EditComplete Gdiff :execute s:Diff('',<bang>0,<f-args>)")
 call s:command("-bang -bar -nargs=* -complete=customlist,s:EditComplete Gvdiff :execute s:Diff('keepalt vert ',<bang>0,<f-args>)")
 call s:command("-bang -bar -nargs=* -complete=customlist,s:EditComplete Gsdiff :execute s:Diff('keepalt ',<bang>0,<f-args>)")
+call s:command("-bang -bar -nargs=* -complete=customlist,s:EditComplete Gtdiff :execute tabpagenr() - 1 . 'tab split' | execute s:Diff('',<bang>0,<f-args>)")
+call s:command("-bang -bar -nargs=* -complete=customlist,s:EditComplete Gtvdiff :execute tabpagenr() - 1 . 'tab split' | execute s:Diff('keepalt vert ',<bang>0,<f-args>)")
+call s:command("-bang -bar -nargs=* -complete=customlist,s:EditComplete Gtsdiff :execute tabpagenr() - 1 . 'tab split' | execute s:Diff('keepalt ',<bang>0,<f-args>)")
 
 augroup fugitive_diff
   autocmd!
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -192,6 +192,16 @@ that are part of Git repositories).
                                                 *fugitive-:Gvdiff*
 :Gvdiff [revision]      Like |:Gdiff|, but always split vertically.
 
+                                                *fugitive-:Gtdiff*
+:Gtdiff [revision]      Like |:Gdiff|, but open the diff in a new tab before
+                        the current working tab.
+
+                                                *fugitive-:Gtsdiff*
+:Gtsdiff [revision]     Like |:Gtdiff|, but split the new tab horizontally.
+
+                                                *fugitive-:Gtvdiff*
+:Gtvdiff [revision]     Like |:Gtdiff|, but split the new tab vertically.
+
                                                 *fugitive-:Gmove*
 :Gmove {destination}    Wrapper around git-mv that renames the buffer
                         afterward.  The destination is relative to the current


blueyed:
Just for reference, an older branch regarding this: https://github.com/tpope/vim-fugitive/compare/tpope:master...blueyed:gtdiff. Cannot check/compare it myself currently.

wadkar:
So, any chance this will get accepted?

plugin/fugitive.vim
```diff
@@ -2516,6 +2519,9 @@ function! s:BufReadIndex() abort
     nnoremap <buffer> <silent> ds :<C-U>execute <SID>StageDiff('Gsdiff')<CR>
     nnoremap <buffer> <silent> dp :<C-U>execute <SID>StageDiffEdit()<CR>
     nnoremap <buffer> <silent> dv :<C-U>execute <SID>StageDiff('Gvdiff')<CR>
+    nnoremap <buffer> <silent> dt :<C-U>execute <SID>StageDiff('Gtdiff')<CR>
+    nnoremap <buffer> <silent> dts :<C-U>execute <SID>StageDiff('Gtsdiff')<CR>
+    nnoremap <buffer> <silent> dtv :<C-U>execute <SID>StageDiff('Gtvdiff')<CR>
```
blueyed: Does that really work?
Seems to be different from what I / @yssl came up with: https://github.com/tpope/vim-fugitive/compare/tpope:master...blueyed:gtdiff#diff-ce13abb0779d494c65de144c88ac3992R2465

wadkar: Yes. It _really_ works.
I didn't see any need for the extra `StageTabDiff` function.
Pull the changes and check it for yourself.

tpope:
I am squeamish about adding 3 commands (not to mention maps) for a feature I don't quite understand. Based on the maps, might I speculate that the primary use case is opening from `:Gstatus`? If so, wouldn't open `:Gstatus` in a tab of its own and doing the traditional diff operations be an equally easy technique?

wadkar:
One of the primary use case for `:Gdiff` (at least for me) is to selectively index hunks. I would speculate that it would be very useful to issue `:Gtdiff` on a long file while working in a window with multiple files. The other commands `:Gtvdiff` and `:Gtsdiff` simply follow existing command patterns so as to not surprise the user. The mappings in the `:Gstatus` seem to extend the logic.

I am fine without the `:Gstatus` mappings, but I would certainly consider `:Gtdiff` as highly effective in cases such as stash popping, selective indexing, or even squashing.

wadkar:
:haha: now that think about it, `dh` and `ds` are mapped to `:Gsdiff`; I should've probably put in an additional `dth` to be completely faithful to existing commands in `:Gstatus`

Anyhow, @tpope does this PR seems reasonable now?

tpope:
Kill the maps for now and squash to one commit. I plan on holding of merging until the next release is cut.

wadkar:
Done. BTW, this is my first _open_ open source contribution.
Thanks, I couldn't have found a better place than `vim-fugitive`

tpope:
congrats :)

wadkar:
oops. I want to rebase my PR on top of your current `tpope/master`
what's the magic incantation for this? (my local `feature/diff_new_tab` is up-to-date with `github/wadkar/feature/diff_new_tab`)

glittershark:
`git rebase 1c84800 --onto tpope/master` should do the trick

glittershark:
(because `1c84800` is the commit immediately before the first commit you want to keep)

xaviergmail:
Any updates on this?

mortona42:
I hope this can get resolved. For working on a laptop, I'd like to see the diff all in one tab, with no other windows. If this can be achieved another way, I'd like to hear about. The proposal in this PR sounded really good to me though.

wadkar:
@tpope I think its high time we merge this into the master ;-)
Remember, this was my first open source contribution - I am not going to take no for an answer 😝 
Just rebased it off master, squashed into one single commit with the same old commit message describing the feature.
Accept please 🙏

wadkar:
Also, revisiting [my old comment](https://github.com/tpope/vim-fugitive/pull/715#issuecomment-159099862) on this thread, I realized it may be more consistent to add `dt` map in the `Gstatus` window. I am still ambivalent about it as it's wasn't my primary use case back then. But from a UX/least surprise perspective, it seems logical to add the following one liner in `s:BufReadIndex()` function (around line 2588):
```vimscript
nnoremap <buffer> <silent> dt :<C-U>execute <SID>StageDiff('Gtdiff')<CR>
```
The commit message does allude to it BTW.

whitebyte:
@tpope still relevant, any chance for it to being merged?

tpope:
I can't remember if I got cold feet or just forgot. I think a 2019 solution would be to put this behavior on `:tab Gdiff`, etc, using `<mods>`. This would sidestep my concerns about bloating the interface with a bunch of new commands.

PR #718, opened 2015-11-18
Adds the U checkout shortcut in :Gstatus to docs

camilopayan:
Just learned about the U shortcut for `git checkout -- [file]` by going on SO, but it wasn't in the docs. So, you know, here.


wadkar:
You might wanna close this PR after looking at #714, which is more comprehensive than this PR.

camilopayan:
Mmm good point @wadkar

PR #729, opened 2015-12-08
Avoid the hit-enter prompt on :Ggrep

BohrShaw:


tpope:
`:Ggrep` is `:grep` extended to `git grep`, full stop. It's not open to bike shedding.

PR #730, opened 2015-12-09
Fix :Gbrowse in "detached HEAD" state

### LINKED ISSUE #646: :Gbrowse doesn't do the right thing when on a detached HEAD.
If I've checked out a remote branch, it should take me the the corresponding file in that remote branch, rather than 

It when I've checked out ``origin/master``, it should take me to
``https://github.com/ClusterHQ/build.clusterhq.com/blob/master/buildbot.tac``
instead of 
``https://github.com/ClusterHQ/build.clusterhq.com/blob//buildbot.tac``
(not the double '//` there).

junegunn:
Closes #646.

This patch tries to find the right URL for HEAD when in detached HEAD state. If there's a matching tag on HEAD (e.g. `git checkout v2.2`) it will use it, otherwise, SHA. This is my attempt to fix the issue without fully understanding the codebase, so maybe I'm doing something incredibly wrong here and if there's a better way to handle this, feel free to close this PR. Actually I couldn't figure out what `^\d\=$` is supposed to match, but I noticed that it matches empty string in detached HEAD state.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -2330,10 +2330,16 @@ function! s:github_url(opts, ...) abort
   if a:opts.revision =~# '^[[:alnum:]._-]\+:'
     let commit = matchstr(a:opts.revision,'^[^:]*')
   elseif a:opts.commit =~# '^\d\=$'
-    let local = matchstr(a:opts.repo.head_ref(),'\<refs/heads/\zs.*')
-    let commit = a:opts.repo.git_chomp('config','branch.'.local.'.merge')[11:-1]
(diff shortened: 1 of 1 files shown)


tpope:
This code is slated to be replaced by rhubarb.vim going forward. I've backported all changes from there which appears to solve the problem.

I'm not sure how I feel about the tag matching behavior. It's definitely useful but I think I might prefer something a bit more explicit. One thing I can say for sure is that it needs to live in `s:Browse()` instead of being GitHub specific, if you are interested in pursuing further.

Issue #376, opened 2013-08-25
Improve startup time


ipkiss42:
With the current implementation, all the code lives in plugin/fugitive.vim. It means that everything is parsed when vim starts, no matter if fugitive is going to be used or not.

Why not moving most of the code to the autoload folder (see :h autoload)? It would improve vim startup time, without any drawback.

lucc:
I am also interested in this. But when I had a look at the code I noticed that there are no comments in the code and it is near to impossible to understand which function does what without reading all the code.

Is there a plan to add comments to the code (like python doecstrings)? Or will PRs for such be merged?

I think this could be a step to get people (me?) interested in doing the autoload implementation.

tpope:
It's an internal API. Documentation is just more work to maintain with very little benefit.

The main section you have to worry about is "Initialization", along with whatever functions it calls from "Utility".

tpope:
Done.

PR #734, opened 2015-12-22
Move most code to the autoload directory.

lucc:
See #376.

The code was moved with minimal changes.  The result according to `--startuptime` if no file under git is opened:

```
old:313.497  005.805  005.805: sourcing /home/luc/.local/share/vim/bundle/vim-fugitive/plugin/fugitive.vim
new:304.365  000.131  000.131: sourcing /home/luc/.local/share/vim/bundle/vim-fugitive/plugin/fugitive.vim
```

**Please test this PR with all the different "entry points" to fugitive.  I do not know and use all of them so I might not have detected some breakage which might be due to this move.**


tpope:
I'm confused by your results as I would expect the `VimEnter *` autocommand to trigger a fugitive load. At any rate I think we need a `FugitiveDetect()` in the plugin file to cope with basically every other case.

lucc:
I see. I did only test a vim start with no file, not a start with a file that is not under git. Sorry, I will try to update the PR.

lucc:
I rebased my changes on the latest master. But now I don't see the problem you mentioned any more. Do you still get it?

I tried to open an empty buffer or a file not under git -> no commands defined
I opened a file under git from the command line and in vim -> commands defined and *seem* to work (only tested some)

justinmk:
@lucc With this functionality (which still exists in your PR):

    autocmd VimEnter * if expand('<amatch>')==''|call fugitive#detect(getcwd())|endif

the move to autoload is almost pointless, it will autoload fugitive (don't know why you aren't seeing that, but a call to `fugitive#foo` will source `autoload/fugitive.vim`). 

One would need to move detection to `plugin/fugitive.vim` (the `FugitiveDetect()` @tpope mentioned) to avoid autoloading the entire thing.  The `BufNewFile,BufReadPost` and other handlers also call `fugitive#detect` so again in every case fugitive will be loaded. Those `fugitive#detect` calls need to be replaced with `FugitiveDetect()`.

neersighted:
I really would appreciate this change as fugitive is a big chunk of my startuptime as well. @lucc, the change is pretty simple, some of the code cannot be autoloaded as it runs on every file load. If you don't have time, I'd be glad to PR the changes to your branch. Let me know.

lucc:
@neersighted sorry I had other stuff to do. If you have the time and send a PR for my branch I will merge it.

neersighted:
@lucc Just wanted to note (two years later, hah), that this work appears to have been done in master, and this PR is now obsolete.

lucc:
@neersighted thanks for the heads up.

PR #741, opened 2016-01-14
Resolve symbolic links in BufNewFile/BufReadPost

pgdouyon:
The symbolic link of an git file has not been resolved by Vim by the time the BufReadPost autocommand is triggered.  Not sure if it's a bug in Vim or not, but attempting to open a version controlled file via a symlink (which is not in a git directory) fails to setup the Fugitive commands in the file.

Basically given the following symlink where $HOME/dotfiles is under version control:
$HOME/.bashrc -> $HOME/dotfiles/bashrc

`:e ~/dotfiles/bashrc` Fugitive setup succeeds
`:e ~/.bashrc` Fugitive setup fails

I realize that it's not a common use case to open git files via symbolic links, but I noticed it when trying to open up some of my dotfiles using bash aliases that use the $XDG_CONFIG_HOME path instead of the git path.

Let me know if you'd rather have the resolve call on the BufReadPost autocommand. I originally had it there, but thought it would be better off closer to the source.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -130,7 +130,7 @@ function! fugitive#extract_git_dir(path) abort
   if s:shellslash(a:path) =~# '^fugitive://.*//'
     return matchstr(s:shellslash(a:path), '\C^fugitive://\zs.\{-\}\ze//')
   endif
-  let root = s:shellslash(simplify(fnamemodify(a:path, ':p:s?[\/]$??')))
+  let root = s:shellslash(simplify(resolve(fnamemodify(a:path, ':p:s?[\/]$??'))))
   let previous = ""
   while root !=# previous
     if root =~# '\v^//%([^/]+/?)?$'


tpope:
You can't call Git commands like `git add` on a symlink either. Why is this Fugitive's problem?

pgdouyon:
Hmm sorry I think I see the confusion and my example was too minimalistic.  It might also be that I don't understand symlinks well enough but if I open a file that lives in a symlinked directory Vim resolves the symlink, but if I open a file that is itself a symlink then Vim doesn't resolve it.

So in Vim:
`:e ~/.bashrc` results in `bufname("%") = ~/.bashrc`
But if the directory `~/.config` is a symlink to `~/dotfiles/config` then:
`:e ~/.config/git/config` results in `bufname("%") = ~/dotfiles/config/git/config`

Not sure if this is expected due to the way symlinks work, but I would expect that a file opened through a symlink directory that's been resolved by Vim would also be resolved by Fugitive.

pgdouyon:
Ughh sorry I thought I tested this better, but I was either out of it yesterday or my environment was borked.

Here's the actual problem I'm having:
`vim ~/.config/git/config` resolves properly
`vim $XDG_CONFIG_HOME/git/config` does not

Guessing the problem has to do with when the environment variable is expanded.  Sorry for all the noise.

pgdouyon:
In either case, this PR is clearly not the right solution since it would also resolve `~/.bashrc` even though it hasn't been resolved by Vim yet.  I guess the problem is that Vim expands XDG_CONFIG_HOME but doesn't resolve it before firing BufReadPost.  

Maybe a better solution is to add another `fugitive#detect(expand('%:p'))` on VimEnter instead?

tpope:
`$XDG_CONFIG_HOME` is resolved by the shell. If it contains `$HOME/.config`, I don't see how this is possible.

pgdouyon:
Can you not reproduce it?  It happens pretty consistently for me.

tpope:
I don't have `XDG_CONFIG_HOME` set, nor do I know what value you have it set to, so no, I have not bothered even trying to reproduce.

pgdouyon:
Here's the commit where it breaks:

https://github.com/tpope/vim-fugitive/commit/0b43b51d7785aeb4002b45ca49cea5aef0d2e988

Before this commit `vim $XDG_CONFIG_HOME/git/config` works, after this commit it doesn't.  Would it be possible to just revert this commit?  Sorry again for all the noise, I thought I had nailed this down better.

pgdouyon:
Steps to reproduce using a local copy of the vim-fugitive repo:
```bash
TEST=~/test
ln -s <localpath>/vim-fugitive/plugin $TEST
vim $TEST/fugitive.vim
:Gstatus
```

Looks like it only occurs when the symlink is to a subfolder in the directory, doesn't happen when the symlink is to the project folder.

tpope:
@blueyed, thoughts?

blueyed:
@pgdouyon 
I also cannot see how `vim ~/.config/git/config` should be different from `vim $XDG_CONFIG_HOME/git/config` - given that it points at the same file.  Using `echo` instead of `vim` should result in the same output, no?  What is `XDG_CONFIG_HOME` set to?
The same applies to your test code: `$TEST` is resolved by the shell and the command would be `vim /home/user/test/fugitive.vim`.
I guess using `resolve()` in the autocommand improves it for you? (`autocmd BufNewFile,BufReadPost * call fugitive#detect(resolve(expand('%:p'))`)

pgdouyon:
I'm guessing the difference is in the value of `<amatch>` vs `bufname("%")`.  I agree there shouldn't be a difference, but at least on my machine there seems to be one.

Playing with it a bit more, it doesn't seem to even require the environment variable.  Both `vim $TEST/fugitive.vim` and `vim ~/test/fugitive.vim` fail to setup the fugitive commands, but accessing it as `cd ~; vim test/fugitive.vim` succeeds.

Before that commit all three variations succeed in setting up the fugitive commands.  To be honest, I have no idea what's going on either and I could only assume that it's a bug or peculiarity in Vim.  Are you able to reproduce the bug using the steps I posted above?

pgdouyon:
Also to clarify, I don't think this PR is the solution since it would cause Fugitive to resolve symlinks that Vim itself hasn't resolved (as in the case of ~/.bashrc).  So I don't really think adding the `resolve()` call is a good solution anymore.

I think I was also careless in how I tried to isolate and test this bug using $XDG_CONFIG_HOME so I think it might be easier to ignore those comments since they're pretty all over the place and a little inaccurate.

I think the example using the vim-fugitive repo is probably simple enough and doesn't depend on anyone's environment (I hope).  Additionally, as I posted above, the bug doesn't even seem to need the environment variable as I originally thought (again due to my careless testing earlier) and seems to actually depend on absolute vs relative paths involving symlink directories.

pgdouyon:
@tpope feel free to close this PR, I'm not really interested in pursuing this any further.  The commit should be simple enough to revert if it comes up again but I'm guessing the original behavior I was relying on was more of an accident/quirk of Vim than anything else.

It doesn't seem to be bothering anyone else as I haven't seen it come up on the issue tracker and I can just update my shell aliases to compensate.  Sorry for the trouble.

PR #750, opened 2016-02-15
added support for gitlab url

serialdoom:


glittershark:
https://github.com/shumphrey/fugitive-gitlab.vim

tpope:
Fugitive is not going to support individual hosting sites going forward. Existing GitHub support is deprecated in favor of rhubarb.vim.

PR #760, opened 2016-03-05
Don't use spaces in Git command to avoid problems under Windows

vadz:
When using a helper script to make Windows Vim work with Cygwin Git, arguments
containing spaces don't survive being passed through "cmd /c" to this script
and are decomposed into several tokens.

Just use "%x20" instead of spaces in the pretty format to avoid the problem.

---
I realize that this is an ugly fix and there may be some bug somewhere among `cmd.exe` invoking Cygwin shell to run my git wrapper script, but this is the only solution to make it work I could find after spending close to an hour on this, so I'd like to submit it just in case it can help somebody else.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -2689,7 +2689,7 @@ function! s:BufReadObject() abort
         if b:fugitive_display_format
           call s:ReplaceCmd(s:repo().git_command('cat-file',b:fugitive_type,hash))
         else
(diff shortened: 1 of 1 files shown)


tpope:
Do you happen to know when support for `%x` was added?

tpope:
I have an old 1.7.1 install it appears to work on so I guess that's good enough.

PR #767, opened 2016-03-18
Add Ggrep[!] option in documentation.

jreybert:
Life would have been much better if I found it sooner!


tpope:
I just did my own take without the comma splice. :smile:

PR #771, opened 2016-03-23
Trigger BufWritePost after adding to index with Gwrite.

airblade:
See airblade/vim-gitgutter#278.


glittershark:
:0

tpope:
Needs a `silent` for the case no `BufWritePost` is defined.

airblade:
Added a `silent`.

airblade:
Thanks for your help with this.

PR #775, opened 2016-03-31
worktree: warn the user that their worktree is broken

### LINKED ISSUE #768: Worktree failures
So I've finally updated to the worktree support that got merged, but using a worktree just outputs:

```
Error detected while processing function <SNR>106_BufReadIndex:                                                                                                                                                                                
line   33:                                                                                                                                                                                                                                     
E344: Can't find directory "`=s:repo().tree()`" in cdpath                                                                                                                                                                                      
E472: Command failed                                                                                                                                                                                                                           
Press ENTER or type command to continue
```

Not sure what's going wrong yet.

mathstuf:
---

FIxes #768. @blast-hardcheese

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -272,6 +272,7 @@ function! s:configured_tree(git_dir) abort
     elseif filereadable(a:git_dir . '/gitdir')
       let worktree = fnamemodify(readfile(a:git_dir . '/gitdir')[0], ':h')
       if worktree ==# '.'
+        echo "Found a broken worktree. Please fix ".readfile(a:git_dir)."/gitdir to contain an absolute path to the worktree's .git file."
         unlet! worktree
       endif
     endif


blast-hardcheese:
:+1: This is exactly where my investigations led me

mathstuf:
@tpope Any thoughts on this?

mathstuf:
I think Git has been fixed for a long time at this point. This probably isn't necessary anymore, but IMO, it'd still be nice to have since the behavior is very confusing, but easy for fugitive diagnose.

tpope:
Should be handled properly by Fugitive now.

mathstuf:
It is indeed handled properly (there's nothing to actually *do* in this case since the `gitdir` file is just wrong). However, *detecting* that this is the case given that you're using a Git release with the broken behavior (though at this point is "ancient", might not be so for users of older, more "stable" distros). This warning message does call out that something is wrong though and, to me, was invaluable in actually finding problematic worktrees.

PR #778, opened 2016-04-06
Use "+ instead of "*

aripollak:
I'm guessing most people will expect "copy to clipboard" to mean the real clipboard, not the primary selection, when working under X11. I certainly did.

From the vim docs:

> *quoteplus* *quote+*
>
>There are three documented X11 selections: `PRIMARY`, `SECONDARY`, and `CLIPBOARD`. `CLIPBOARD` is typically used in X11 applications for copy/paste operations (`Ctrl-c`/`v`), while `PRIMARY` is used for the last selected text, which is generally inserted with the middle mouse button.
>
> Vim's X11 clipboard providers only utilize the `PRIMARY` and `CLIPBOARD` selections, used for the '*' and '+' registers, respectively.


tpope:
Not sure I like it, but you have a point so I'm merging.

PR #785, opened 2016-04-20
Fix for empty buffer :Git command in Neovim

### LINKED ISSUE #783: :Git don't play well with empty buffers
:Git status

Error detected while processing function <SNR>17_Git:
line   11:
E499: Empty file name for '%' or '#', only works with ":p:h":     tabedit %

screen shot 2016-04-13 at 5 34 13 pm

J3RN:
Detects whether the current buffer is empty; opens a new empty tab if so, a new tab of the same buffer if not.

Fixes #783

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -700,7 +700,11 @@ function! s:Git(bang, args) abort
   let args = matchstr(a:args,'\v\C.{-}%($|\\@<!%(\\\\)*\|)@=')
   if exists(':terminal')
     let dir = s:repo().tree()
-    -tabedit %
+    if expand('%') != ''
+      -tabedit %
+    else
+      -tabnew
+    endif
     execute 'lcd' fnameescape(dir)
     execute 'terminal' git args
   else


tpope:
This breaks commands like `:Git log %`

J3RN:
I suppose that explains the motivation for using `tabedit`. Alright, this is a little bit longer, but what about something like this:

    if expand('%') != ''
      -tabedit %
    else
      -tabnew
    endif

Pretty much an explicit workaround, but it's the best I could come up with offhand.

eloytoro:
Where are we on this? I can verify this fixes the issue for me as well

justinmk:
The commit message and subject line are rambling; that could be improved. The code change could also be a single `exe` line.

J3RN:
@justinmk I'm not exactly a seasoned Vim-scripter. What is the `exe

plugin/fugitive.vim
```diff
@@ -700,7 +700,11 @@ function! s:Git(bang, args) abort
   let args = matchstr(a:args,'\v\C.{-}%($|\\@<!%(\\\\)*\|)@=')
   if exists(':terminal')
     let dir = s:repo().tree()
-    -tabedit %
+    if expand('%') != ''
+      -tabedit %
+    else
+      -tabnew
+    endif
```
justinmk: execute @% == '' ? '-tabnew' : '-tabedit %'

yevhen-m:
What's the status of this issue? Is the proposed fix going to be merged?

yevhen-m:
@J3RN, @tpope thanks.

PR #786, opened 2016-04-28
Gstatus: Map cF to `Gcommit --fixup=HEAD`.

justinmk:


doc/fugitive.txt
```diff
@@ -42,6 +42,7 @@ that are part of Git repositories).
                         -     |:Git| reset (staged files)
                         cA    |:Gcommit| --amend --reuse-message=HEAD
                         ca    |:Gcommit| --amend
+                        cF    |:Gcommit| --fixup=HEAD
```
justinmk: In analogy to `ca`/`cA` ideally we would provide a `cf` mapping that pre-filled the command:

    :Gcommit --fixup=

where `<tab>` would invoke auto-completion of branch names or ancestors of HEAD. But `cF` probably covers 95% of use-cases.

justinmk:
This might not a sufficiently generic workflow, so closing.

Issue #753, opened 2016-02-25
:Ggrep should not run from top-level git dir


dmerejkowsky:
Use case: I have a project with some documentation in doc/, some Python code in python/, and some CMake code in cmake/

If I run :Ggrep, fugitive will set working dir to the top-level git dir, and I'll get matches everywhere, when I only want results in my current working dir.

Update: the culprit seems to be here:

``` VimL
function! s:Grep(cmd,bang,arg) abort

 " ...
  try
    execute cd.'`=s:repo().tree()`'
    " ...

   finally
      execute cd.'`=dir`'
   endtry
endfunction
```

I'm using version 2.2. 

I've tried commenting the lines running `execute cd`, but then fugitive calls git with

```
git grep --git-dir /path/to/proj/.git foo
```

which cannot work (--work-tree should be set to /path/to/proj for this to work)

tpope:
See #787.

PR #787, opened 2016-04-30
Ggrep: use current working directory

dmerejkowsky:
Also, always call git with
 --git-dir and
 --work-tree

Fixes issue #753

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -378,7 +378,11 @@ endfunction
 call s:add_methods('repo',['dir','tree','bare','translate','head'])
 
 function! s:repo_git_command(...) dict abort
-  let git = g:fugitive_git_executable . ' --git-dir='.s:shellesc(self.git_dir)
+  let git_dir = s:shellesc(self.git_dir)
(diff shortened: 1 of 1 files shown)


tpope:
So what happens if the working directory isn't under the work tree at all?

dmerejkowsky:
Hum. Did not think of this use case. 

Let's assume you have a repo in `/path/to/repo` , and `/path/to/repo/bar/baz` that matches `foo`

If you run

```
:e /path/to/repo/some-file
:Ggrep foo
```

you get a quickfix list containing `bar/baz` : it's relative to the git repo and so you cannot open the file directly from the quickfix list.

Without my patch, you do get `/path/to/repo/bar/baz` in the quickfix list.

tpope:
As a compromise, `:Ggrep pattern .` now works to target a subdirectory. I don't think doing more than that is tenable.

dmerejkowsky:
> As a compromise, :Ggrep pattern . now works to target a subdirectory. I don't think doing more than that is tenable.

Excellent idea, I like it :)

PR #788, opened 2016-05-03
Add Gtcd command

qstrahl:
`:tcd` was introduced to neovim on August 24th, 2015; this wraps it with `:Gtcd`. I didn't think it was necessary to check if `:tcd` exists, but if you think otherwise, I'll make it conditional.

### DIFF
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -29,6 +29,9 @@ that are part of Git repositories).
                                                 *fugitive-:Glcd*
 :Glcd [directory]       |:lcd| relative to the repository.
 
+                                                *fugitive-:Gtcd*
+:Gtcd [directory]       |:tcd| relative to the repository. (Neovim only)
+
                                                 *fugitive-:Gstatus*
 :Gstatus                Bring up the output of git-status in the preview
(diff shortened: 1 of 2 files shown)


nicolaiskogheim:
Many vim users are oblivious of neovim, and likewise about the `:tcd` command. I suggest you mention it in the docs that this is a neovim-only feature. Moreover, I think its wise to check if `:tcd` exists; we wouldn't want people filing a bug report here every time `:Gtcd` throws an error in vim.

qstrahl:
@nicolaiskogheim Thanks for the advice! Took me a while to find the time, but I made the changes. Hopefully this is harmless enough now.

qstrahl:
Anything I can do to help push this along?

qstrahl:
I've lost interest in this and I was quite probably the only user to ever have any interest in this, so I'm closing it

tpope:
I have a massive backlog from the last half of 2016 I will probably never fully work through. I am wary of `:tcd` because I don't understand how it interacts with `haslocaldir()` and `:lcd` in general. That said I would entertain adding `:Gtcd` as an undocumented feature.

PR #801, opened 2016-06-01
Add shortcuts for getting hunks from other views of the diff.

### LINKED ISSUE #798: mapping for ``diffget //2`` and ``diffget //3``?
I really like `dp` and `do` shortcuts in two-split merge window. I know that `dp` works even in the three-split merge window, but for `diffget` I have to run whole complicated `:diffget //2<CR>` and `:diffget //3<CR>`. Could we get some Gdiff (or diff) specific shortcuts, like e.g.

```
nmap d2 :diffget //2<Cr>
nmap d3 :diffget //3<Cr>
```

Of course I can add these mappings to my `~/.vimrc`, but it would be lovely if they were just three-split window specific.

mcepl:
Fixes #798


mcepl:
ping?

mcepl:
@tpope why do you hate me?

justinmk:
@mcepl isn't this missing documentation?

mcepl:
OK, I will write something down in the next PR. Thanks, @tpope for mering it.

PR #805, opened 2016-06-12
Gsdiff: split horizontally with diffopt+=vertical

blueyed:
With `set diffopt+=vertical` `Gsdiff` will not use a horizontal split,
because the effective command being run is `:leftabove keepalt diffsplit`.

This patch temporarily unsets the `vertical` option.

~~If this is considered to be good, then this should be done for the 3-way-diff, too.~~ (that does not use `diffsplit`, so it should be fine)

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1858,6 +1858,10 @@ function! s:Diff(vert,keepfocus,...) abort
       setlocal cursorbind
     endif
     let w:fugitive_diff_restore = restore
+    if !empty(a:vert) && &diffopt =~# 'vertical' && a:vert !~# ' vert '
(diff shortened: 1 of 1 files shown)


tpope:
Damn, missed this, ended up implementing it myself.

blueyed:
With c972b2a it does not honor `diffsplit+=vertical` with `:Gdiff` anymore - i.e. uses horizontal there.

blueyed:
btw: `fugitive-:Gsdiff` is still in the help although it does not exist anymore.  Might be intentional though?

blueyed:
updated comment above

tpope:
Yes, intentional, so people can find the replacement.

Issue #147, opened 2012-01-27
Follow symlinks (and act on the target repository)


blueyed:
If I edit ~/.pentadactylrc (not in Git) which is a symlink into ~/.dotfiles/pentadactylrc (in Git) I would like `:Gcommit` et al to work on it, too.

This should be possible if fugitive would follow symlinks and recognize that the file itself is located in a git repository and add appropriate parameters to invocations of git, i.e. '--git-dir'.

blueyed:
@wilywampa 
You can use `call fugitive#detect(resolvedfile)` instead.

I've added it to my method: https://github.com/blueyed/dotfiles/blob/ab2850675bbfcf0db18dbf81a31b90e65aaae7f8/vimrc#L1247-L1278

AssailantLF:
I made a Windows compatible version of the MyFollowSymlink function. It's really messy and inefficient, and needs to be ran from the buffer of the symbolic file, but seems to work for simple cases. Hopefully others may find it useful, or can improve upon it:

    function! MyFollowSymlink(...)
      let lit = a:0 ? a:1 : '%:p'
      let fpath = expand(lit)
      let recur = a:0 > 1 ? a:2 : 10
      if !recur
        return
      endif
      let matstr = '<SYMLINK..\s\+' . expand(lit.':t') . ' ['
      " check if it's a symlink
      if (match(system('dir ' . fpath . '*'), matstr) == -1)
        " check if parent dir is symlink
        call MyFollowSymlink(lit.':h', recur-1)
      else
        " extract symlink path
        let substr = '.*<SYMLINK..\s\+' . expand(lit.':t') . ' \[\(.\{-}\)\].*'
        let fpath = substitute(system('dir ' . fpath . '*'), substr, '\1', "")
        " figure out current file's path
        let resolvedfile = filereadable(fpath) ?
              \fpath : findfile(expand('%:t'), fpath, '**')
        exec 'file ' . resolvedfile . ' | w! | doautocmd BufReadPost'
      endif
    endfunction

burgerga:
It's quite confusing because gitgutter does follow the symlink, so I can see the changes but not act on them ;)

dalejung:
@burgerga gitgutter handles the symlink internally. i just patched my local vim-fugitive to do the same and it fixes my use case. https://github.com/dalejung/vim-fugitive/commit/21b91d4c18dcdf2a6292a172ab09776d41a6b715 if you want to try.

burgerga:
[...]

PR #814, opened 2016-06-29
Follow symlinks

redbmk:
Addresses issue #147
Works with vim-airline as well

@tpope, please check this out and see if you can break it. It's working fine for me, including with deep symlinks as @stardiviner suggested. An extra patch would likely be required for windows, but I have no easy way of testing that so this only fixes *nix systems. As far as I can tell, it shouldn't break the windows version (unless `resolve` does something weird on windows).

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -140,12 +140,13 @@ function! fugitive#extract_git_dir(path) abort
   if s:shellslash(a:path) =~# '^fugitive://.*//'
     return matchstr(s:shellslash(a:path), '\C^fugitive://\zs.\{-\}\ze//')
   endif
+  let path = resolve(fnamemodify(a:path, ':p'))
   if isdirectory(a:path)
-    let path = fnamemodify(a:path, ':p:s?[\/]$??')
+    let path = fnamemodify(path, ':p:s?[\/]$??')
   else
-    let path = fnamemodify(a:path, ':p:h:s?[\/]$??')
+    let path = fnamemodify(path, ':p:h:s?[\/]$??')
   endif
-  let root = s:shellslash(resolve(path))
+  let root = s:shellslash(path)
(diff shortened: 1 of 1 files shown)


HaleTom:
I'd love to see this merged in. @tpope do you have any comments to be actioned when resolving the merge conflicts?

redbmk:
I fixed the merge conflicts. @HaleTom thanks for pointing those out :)

redbmk:
@idbrii Good call. I wasn't sure whether to rebase or merge. That's fixed now so it's just one commit on top of `master`.

plugin/fugitive.vim
```diff
@@ -140,10 +140,11 @@ function! fugitive#extract_git_dir(path) abort
...
   endif
+  let path = resolve(fnamemodify(a:path, ':p'))
   if isdirectory(a:path)
-    let path = fnamemodify(a:path, ':p:s?[\/]$??')
+    let path = fnamemodify(path, ':p:s?[\/]$??')
   else
-    let path = fnamemodify(a:path, ':p:h:s?[\/]$??')
+    let path = fnamemodify(path, ':p:h:s?[\/]$??')
   endif
   let root = s:shellslash(resolve(path))
```
idbrii: Does adding the above resolve do anything? There's one here too.

redbmk: The `fnamemodify` gets the directory of the file. Without the `resolve` above, this will get the path of the symlink, not the path of the target. Now that you mention it though, I think this second `resolve` is unnecessary.

mk12:
Any update on this? I'd also love to see this merged.

mk12:
Thanks for the quick reply!

That makes a lot of sense. I think I'll just create a shortcut for your Resolve command and do it manually when I need to, which isn't too often.

PR #817, opened 2016-07-09
Allow paragraph motion in gitcommit buffers

tweekmonster:
Since gitcommit buffer lines are pretty much all comments, it would be
sensible to treat lines that only has a comment charater as blank.


tpope:
Why does the first line end with `|`? Am I missing something?

tweekmonster:
Crap, that was a leftover.  I originally had it set in a `FileType` autocmd but the braces were breaking the syntax highlighting.  I can fix it real quick.

tpope:
Should be moot for the new `:Gstatus`.

PR #823, opened 2016-08-07
Fix U does not delete Untracked files

### LINKED ISSUE #765: U does not delete Untracked files
Is it possible to `rm` them? Thank you.

tmsanrinsha:
This fixes #765 .
But the unworkable message "To restore, :Git cat-file blob fatal:  > foo" remains.


tpope:
What do you mean unworkable?

tmsanrinsha:
> :Git cat-file blob fatal: > foo

I tried this command but it didn't work, because blob is fatal:.

tpope:
Weird, hopefully someone can look into that.

Issue #582, opened 2014-11-25
Show :Gstatus at full-width top for vertical diff splits


veger:
I have the same issue as described http://stackoverflow.com/q/24407556/246263 (it has nice ASCII art to visualize the issue).

I use ```set diffopt=vertical``` in vimrc to get the vertical 3-way diff, without this option the 3-way diff is horizontally split.
But, if I open the status window using ```:Gstatus``` it opens above the active split window only, instead of full-width at the top.

Is it possible to change this behavior, or add an option to (better) support vertical diffs?

Note that in the vimcast (http://vimcasts.org/episodes/fugitive-vim-working-with-the-git-index/) vertical splits (magically?) does seem to work as expected...

atrioom:
Yes, exactly that. My `:Gdiff` command also opens the windows horizontally split. Searching for a fix, this issue here is the only thing I found...  So +1 from me.

tpope:
Just use `:Gvdiff`.

redbmk:
I don't see how that solves the issue. In a 3-way merge, if I have a single file open and type `:Gvdiff`, it opens all three buffers side-by-side. Then I type `:Gstatus` and the status bar goes above the current buffer like described in the stack overflow issue, not over all three like it does in the vimcast.

I fixed this in my vimrc though by creating a custom mapping:

`nmap <leader>gs :Gstatus<cr><c-w>k<c-w>K<c-w>p`

tpope:
I misunderstood. I think a well placed `exe &splitbelow ? 'rightbelow' : 'leftabove' ...` would fix it. Feel free to send a pull request.

redbmk:
I played around with `rightbelow` and `leftabove` but it didn't seem to make a difference. In my own vimrc I had found out that `<c-w>K` after calling `:Gstatus` gave me the behavior described here (the window gets the full width).

If that's not the desired default behavior maybe adding a setting that can be changed would be better. Just let me know.

redbmk:
@justinmk pointed out that this didn't work very well when using `noequalalways`. I did a little more digging and found that instead of `rightbelow` and `leftabove`, we could use `botright` and `topleft` to achieve the full-width affect without messing up window layouts when using `noequalalways`.

PR #833, opened 2016-09-08
Show :Gstatus at full-width top for vertical diff splits

redbmk:
Fixes issue #582

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -753,6 +753,7 @@ function! s:Status() abort
   try
     Gpedit :
     wincmd P
+    wincmd K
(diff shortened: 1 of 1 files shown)


justinmk:
Did you test this with 'noequalalways' ?

redbmk:
I have now - hadn't heard of that before.

I can squash the commits before merging if this gets approved...just figured I'd keep them both there for reference.

With the first fix if you had some horizontal splits it would end up resizing them after you close the window. Doing it this way seems to only move the windows around until you close this window, then puts them back where they were.

tpope:
I implemented something similar alongside the new `:Gstatus`.

Issue #303, opened 2013-02-06
fugitive workflow for commit --fixup?


lkraav:
Wondering first of all if there's a more fluid way to fugitive this besides typing :Gcommit --fixup=<findhashfromsomewhere>

jreybert:
One interesting feature would be to ease the fixup when this is a one line fix (or all modified lines belong to the same commit).

Then, the map could take as --fixup argument the line commit SHA1. With one command, the line fix is commited with the right fixup SHA1.
However, Gblame shows zeros when the line is modified...

lwm:
In the meanwhile, [this is awesome](https://blog.mikecordell.com/vim/2014/07/20/quick-fixup-in-vim-with-fugitive.html).

still-dreaming-1:
If anyone here is still interested in this, I made an implementation of this, which is being discussed in pull request #843 . This would be a good time to provide some feedback and input.

tpope:
Provided as `cf` as part of the new `:Gstatus`.

ultrox:
I love this feature, but it seems I have a problem understanding to find commit. 

In documentation it reads:
```
cf                      Create a `fixup!` commit for the commit under the
                        cursor.
```
So I open :Gclog and do cf there, but nothing happens. Am I been missing something obvious? How can I find a list to `cf` on?

I also tried to `cf` unpushed, and that works fine.

tpope:
> part of the new `:Gstatus`.

tmccombs:
but what if you want to fixup a commit that isn't shown in the status buffer?

tpope:
It works editing a commit object too.

ultrox:
Oooh, now I get it! @tpope you are cryptic fella! :smile: 

Basically, I just did 
1. :Gcommit, found the commit I want, 
2. 'Enter' to it to get it to editing, 
3. `cf` when the cursor was line after committer 

This works fine.

I was expecting to be able to `cf` from QuickFix List.

tpope:
The quickfix window is global. It's not our place to go dumping Fugitive specific maps in it. It's also problematic in other ways. These days I recommend using `:Git log` instead.

ultrox:
Absolutely make sense!

PR #843, opened 2016-10-07
commit --fixup

still-dreaming-1:
This is the commit --fixup workflow requested in issue #303. Let me know if you like the general idea, and if you do, what I need to change before you can merge it.

### DIFF
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -43,6 +43,7 @@ that are part of Git repositories).
                         cA    |:Gcommit| --amend --reuse-message=HEAD
                         ca    |:Gcommit| --amend
                         cc    |:Gcommit|
+                        cf close status and |:GcommitFixup|
                         cva   |:Gcommit| --amend --verbose
                         cvc   |:Gcommit| --verbose
                         D     |:Gdiff|
@@ -73,6 +74,16 @@ that are part of Git repositories).
                         command, it is possible (but unadvisable) to alter the
                         index with commands like git-add and git-reset while a
                         commit message is pending.
+
+                                                *fugitive-:GcommitFixup*
+:GcommitFixup           A wrapper around git-commit --fixup=commitID. Brings
+                        up the git log. First move the cursor to the line with
+                        the commit ID you want to use with --fixup, then press
+                        f. Press n to jump to the next commit, and N to
+                        jump to the previous commit. At the end you get a
+                        chance to merge this fixup commit into the commit
+                        being fixed up. Alternatively, you can run this
+                        command via cf from the |:Gstatus| buffer.
 
                                                 *fugitive-:Gmerge*
 :Gmerge [args]          Calls git-merge and loads errors and conflicted files
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1132,6 +1132,30 @@ function! s:FinishCommit() abort
   return ''
 endfunction
 
+" Section: GcommitFixup
+call s:command("GcommitFixup :call s:CommitFixup()")
+
+function! s:CommitFixup()
+  call s:Git(1, "log")
+  nnoremap <buffer> f :call fugitive#commit_fixup_with_commit_id_from_current_line()<CR>
+  nnoremap <buffer> n :call search('^commit ')<CR>
+  nnoremap <buffer> N :call search('^commit ', 'b')<CR>
+endfunction
+
+function! fugitive#commit_fixup_with_commit_id_from_current_line()
+  normal! ^w
+  let commit_id = expand('<cword>')
+  bdelete
+  call s:Commit('--fixup='.commit_id)
+  call inputsave() " avoid problems with Vim's typeahead mechanism by calling this before input()
+  let rebase_parameters = 'rebase -i --autosquash '.commit_id.'^'
+  let rebase_autosquash = input('Would you also like to perform an interactive autosquash rebase on that commits parent now? (git '.rebase_parameters.')? (y/n):')
+  call inputrestore() " avoid problems with Vim's typeahead mechanism by calling this after input()
+  if rebase_autosquash ==? 'y' " Play it safe. Anything other than 'y' or 'Y' means no.
+    call s:Git(0, rebase_parameters)
+  endif
+endfunction
+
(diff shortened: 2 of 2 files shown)


justinmk:
I did something related in https://github.com/tpope/vim-fugitive/pull/786 but TBH I've been happy with this simple mapping:

    nnoremap UF :Gcommit --fixup=

To fixup head, just enter `@` (git alias for HEAD):

    :Gcommit --fixup=@

otherwise paste whatever hash you just yanked from somewhere, or use fugitive's `<C-R><C-G>` mapping to paste the current object hash.

Given how simple this is I'm not sure it needs to be built-in.

still-dreaming-1:
@justinmk Interesting. Maybe I should change the mapping from C-f to cf. That would be more along the lines of the other mappings. Also, it would be more Vimlike, control should not be used if when there is a better fit out there. If I understand your comments from your pull request, you did not think the workflow was sufficient because you could not tab through different commit ids. Is that right? This brings up the output from `:Git! log`, but it also overrides the enter key mapping so that it selects the commit id from whatever line the cursor is on. So in my opinion, this is even nicer than tabbing through commit ids because you also see the log. Another nice thing is the cursor starts on the line with the commit id of the previous commit, so if you want to use that one, you just press enter.

I think I will add a mapping so that `n` jumps to the next commit. Unless there is already something like that? I'm not sure where the mappings from that buffer are defined. I don't see them from in the fugitive code, but I discovered by accident that pressing enter brings up more details about a commit, that is before I overrode that mapping. But you can still use that mapping if you bring up `:Git! log` yourself. Maybe `f` should be used instead of enter to select the commit so I don't have to override that mapping...Speaking of which, where is the original enter mapping defined in the code that brings up a diff for that commit? Is that even a fugitive thing? How do I go back to the log from there?

bbinet:
I also think `cf` is a better mapping than `C-f` because it will be more consistent to the `cA`, `ca` commit --amend mappings.

bbinet:
Also +1 for using the `f` key to select the commit to fixup, because it could be useful to still be able to see the full commit description when pressing `enter`

bbinet:
When selecting a commit to fixup, it could also be useful to ask the user if Fugitive should also rebase with autosquash (`git rebase -i --autostash commit^`)

still-dreaming-1:
Now I have it so that you press `cf` instead of `Control-f` from the `:Gstatus` buffer. Also, you press `f` instead of `enter` to select a commit id from the log.

@bbinet The auto-rebase option is interesting. I'm not sure if I like it or dislike it... What do other people think?

Before we get too carried away, what does @tpope think of all this? Is there any chance this pull request will get merged at some point?

still-dreaming-1:
Now I got it so that once the log is displayed during this workflow, you press `n` or `N` to jump to the next or previous commit.

still-dreaming-1:
@bbinet After further thought, I agree it would be nice for fugitive to ask you if you want to do an autosquash like that as that is typically why you are doing a fixup commit. That would help you to keep a clean commit history as you work rather than leaving it to cleanup at some future point.

I will do some research on getting Vim to ask the user a question, but I don't believe I have ever written that functionality into a plugin before.

still-dreaming-1:
It is asking the user and doing that autosquash rebase now. I went through the full workflow, imagining I was just hacking on some project and wanting to do that stuff with git. Combined with the rebase it is a pretty sweet workflow. I need to improve the way that last part works as at least on my system it ends up running Neovim inside of Neovim, which somehow did not cause any problems for me...

still-dreaming-1:
I just wanted to summarize the idea behind the complete workflow in case that adds clarity of why the autosquash at the end is good.

You do a commit. Then you do some more work. You realize it is just more of the same from the previous commit, perhaps the previous commit could even be considered "unfinished". So you want to do a fixup commit. No problem, just stash the changes from `:Gstatus`, and then use `cf` to do a fixup commit. It brings up the log so you can select a commit. The first one is what you want so you immediately press `f` to use that as your fixup commit. It asks you if you want to do the rebase. You just want your latest commit to be a part of the previous one, so you are happy this makes the whole process feel like an atomic operation. So you type `y` and press enter. Done. You see the interactive rebase buffer. Everything looks good, so you save and quit (this part needs adjusting), leaving you looking at the original file you were editing.

Done. There is nothing stopping you from doing a more complex interactive rebase later on, if that is your cup of tea, but now that last commit has been cleanly combined as part of the previous one.

bbinet:
@still-dreaming-1 thanks, it looks great!

still-dreaming-1:
So opening Neovim inside of Neovim for the interactive rebase buffer was a result of having nvim set as the default editor for my system. It actually did not cause any problems, it is just a little strange. I'm not sure how to avoid that, am I am open to suggestions, but it was not a big deal for me.

I tried setting vim as the default better and the experience was a little nicer. It opened up a new instance of Vim for that buffer. Once it closed, I saw some output in the terminal, and then it went back to the original Vim instance. I'm not sure why Vim works nicer, but maybe there is something to learn from that to get Neovim to work the same way? I don't know if Vim works differently because of some special handling on the side of fugitive or if that is just a Vim thing.

justinmk:
> I'm not sure why Vim works nicer, but maybe there is something to learn from that to get Neovim to work the same way? 

It's because `:!vim` works in terminal-Vim. `:!` uses pipes so it works consistently in all UIs, but for interactive terminal applications `:terminal` is needed (which fugitive uses for nvim instead of `:!`).

We could actually fake Vim's `:!vim` behavior with `:terminal` (and it would even work in GUIs), so that it works like terminal-Vim's `:!vim`. But it might be a better use of time to figure out specifically what strangeness you observed and see how we can address that.

still-dreaming-1:
@justinmk Well nothing bad happened. But I could see both Neovim's at the same time. I could see Neovim running inside of Neovim. I guess it is ok like that, but I'm not sure.

still-dreaming-1:
There has not seen any activity in the way of comments for a while, and I am happy with it now, so I did a rebase and brought it down to a single commit, which seems right given how little has changed. I also updated the documentation to fix something I forgot to change and, also in the documentation, I explained a little about what the rebase question at the end is for.

PR #845, opened 2016-10-12
Add ssh://git@ remote support

SevereOverfl0w:
This adds support for the SSH git scheme.

A mirror commit should probably make it's way over to Rhubarb.


tpope:
Yes please submit it there, as that support is canonical while this is deprecated.

PR #854, opened 2016-11-05
Add a bit of documentation for d2o and d3o commands.

mcepl:
Covering changes in #801.

### DIFF
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -181,7 +181,9 @@ that are part of Git repositories).
                         to the right or bottom, depending on 'diffopt' and
                         the width of the window relative to 'textwidth'.  Use
                         |do| and |dp| and write to the index file to simulate
-                        "git add --patch".
+                        "git add --patch". For the three-way diff, there is
+                        also d2o and d3o pulling the hunk to the middle from
+                        the left or the right window, respectively.
 
                                                 *fugitive-:Gsdiff*
 :Gsdiff [revision]      Like |:Gdiff|, but always split horizontally.


justinmk:
The documentation should probably live under `:Gblame` next to the other mappings. https://github.com/tpope/vim-fugitive/pull/854/files#diff-2d13dec36bf619dd45b9bf4d1bd631a5R208

mcepl:
> The documentation should probably live under :Gblame next to the other mappings.

I don’t understand. ``Gdiff`` mapping already has documentation above those mappings in the same section of the file. We probably may need to switch this documentation in a separate section if anything, but I don’t understand how moving it couple of lines down helps anything.

justinmk:
Ok, but at least they should not be surround in single-quotes (`'d2o'`), that's reserved for options--not commands or mappings--in Vim docs.

mcepl:
Better?

tpope:
Can you make it one big paragraph please?

doc/fugitive.txt
```diff
@@ -182,6 +182,9 @@ that are part of Git repositories).
                         the width of the window relative to 'textwidth'.  Use
                         |do| and |dp| and write to the index file to simulate
                         "git add --patch".
+                        For the three-way diff, there is also |d2o| and |d3o|
```
justinmk: Bars are for tags, these tags go to nowhere. Perhaps backticks:

    `d2o`

or double quotes:

    "d2o"

mcepl: How is it different from |do| and |dp| just a line above?

mcepl:
OK, I think this is the proper way how to do it.

mcepl:
BTW, ping?

tpope:
Remind me, how did we get from "Can you make it one big paragraph" to a subsection in a completely different part of the file?

mcepl:
I am afraid I lost that during those five months you missed this PR ;).

Sorry, I probably really did not understand what you expected. Could you repeat, please, or elaborate what did you expect to be in that paragraph? Should I rewrite ``:diffget`` help in the form appropriate for ``d2o``?

tpope:
Piecing things together from my emails, the patch originally looked something like
```
@@ -182,6 +182,9 @@ that are part of Git repositories).
                         the width of the window relative to 'textwidth'.  Use
                         |do| and |dp| and write to the index file to simulate
                         "git add --patch".
+                        For the three-way diff, there is also |d2o| and |d3o|
```

All I wanted was for you to take `For the next...` and put it on the same line as `"git add --patch".`. Also @justinmk told you to drop the `|`s because those aren't valid help tags, and that still stands.

mcepl:
> All I wanted was for you to take For the next... and put it on the same line as "git add --patch"..

Aaaaahhh! I see. OK, I can certainly do it.

> Also @justinmk told you to drop the |s because those aren't valid help tags, and that still stands.

And why is it already OK in the same paragraph for ``do`` and ``dp`` commands? What’s the difference?

tpope:
Compare `:help do` and `:help dp` to `:help d2o` and `:help d3o`.

tpope:
Oh, and in accordance with the contribution guidelines, drop the period in your commit subject.

mcepl:
What about this one?

tpope:
How many times are you going to rip the nice simple singular paragraph away from me

mcepl:
Whatever you say, sir!

mcepl:
Thank you and sorry for being such a bother.

Issue #851, opened 2016-10-22
Git worktree and fugitive issues


smani:
**Steps to reproduce:**
1. mkdir temp
2. cd temp/
3. git clone git://github.com/tpope/vim-fugitive.git
4. cd vim-fugitive/
5. git branch wt remotes/origin/worktree
6. git worktree add ../wt wt
7. cd ..
8. cd wt
9. vim plugin/fugitive.vim
10. `:Gstatus`
    - `Error detected while processing function <SNR>18_BufReadIndex:`
    - `line   30:`
    - `E344: Can't find directory "C:/temp/wt/" in cdpath`
    - `E472: Command failed`
11. `:Gblame`
    - `fugitive: file or blob required`

**My setup:**

`Vim:`
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Apr  9 2016 08:08:03)
Included patches: 1-1721

`vim-fugitive:`
 58ed86e434951319ff7d438fb1096f24a51e6458

`Git:`
git version 2.10.0.windows.1

smani:
It doesn't seem to happen on Linux. I see this only on Windows. I use the `vim` within `Git Bash` (MINGW64) terminal.

tpope:
Closed per discussion in #855.

PR #855, opened 2016-11-25
Fix worktree issues in MSYS2/Cygwin environment

smani:
Addresses #851

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -68,6 +68,30 @@ function! s:shellslash(path) abort
   endif
 endfunction
 
+function! s:parse_dotgitfile(fpath) abort
+  let line = get(readfile(a:fpath, '', 1), 0, '')
+  " Git-for-Windows records Windows path names in '<work-dir>/.git' FILE
+  " However, the VIM from Cygwin/MSYS2 fails to 'cd <windows-path>'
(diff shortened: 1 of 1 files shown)


tpope:
So if I understand this correctly, MSYS2 Git has generated a file like `gitdir: C:/path/.git`, and this patch changes it to `/c/path/.git` (or however it is Cygwin does it) so Fugitive in Cygwin Vim can load. But once it *is* loaded, I would expect lots of things to fail, as Fugitive does a lot of stuff like `git --git-dir=/c/path/.git` under the hood. So my question is, how well does (or did) Fugitive work with this in place?

smani:
Thats correct about the path conversion. However, with this fix I didn't find any functionality to be broken. At least Gstatus and Gblame commands were working. The issue #851 has the steps to reproduce the issue this patch addresses. I must confess that I don't use Windows anymore and don't have a setup to reproduce the issue now.

tpope:
In 2016 much of the code base used a temporary `:cd`, so it wouldn't surprise me if those 2 commands worked. Nowadays, those have mostly been replaced with `git -C`, so I would expect this patch to accomplish little.

If anyone with this very particular msys2/cygwin wants to pursue this further, start with a report of which commands work and which don't.

PR #859, opened 2016-12-13
Use %delete_ to delete all the lines in the buffer

### LINKED ISSUE #829: I need to type `:Gwrite` twice in order to add a file
Not sure why this is happening but I consistently have to type `:Gwrite` twice whenever I want to add a file.

edkolev:
This is intended to fix #829 

I couldn't find a bug in the original approach (`$read '.s:fnameescape(file)`, then `'1,'.last.'delete_'`),  I don't see an issue in fugitive's code, maybe there's a bug in my version of vim.

But this alternative implementation is working fine: ` %delete_`, then `read '.s:fnameescape(file).'`, then `1delete_`.

Let me know if the PR can be improved.

Thanks for this plugin

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1577,9 +1577,7 @@ function! s:Write(force,...) abort
         endif
         try
           let lnum = line('.')
-          let last = line('$')
-          silent execute '$read '.s:fnameescape(file)
-          silent execute '1,'.last.'delete_'
+          execute 'silent %delete_|silent read '.s:fnameescape(file).'|silent 1delete_'
           silent execute lnum
           set nomodified
           diffupdate


tpope:
Is this still a problem for you? Your approach seems okay, but I'm generally reticent to merge a solution when I don't understand the problem.

tpope:
More explicit: Confirm you can reproduce the problem with no other plugins loaded and we can talk about merging.

edkolev:
I *can't* reproduce it with this vim version:
```
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Mar 15 2017 15:21:13)
MacOS X (unix) version
Included patches: 1-458
```

But I *can* reproduce it with this slightly older version (and no plugin but fugitive) :
```
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Nov 25 2016 11:41:41)
Included patches: 1-101
```

tpope:
If this is still a problem on the rewritten version, happy to discuss further, but I don't think this particular fix will work with the new generalizations.

tpope:
Should be fixed. I misread this as a fix for `:Gread`, not `:Gwrite`, so my remark about "new generalizations" does not apply.

PR #862, opened 2016-12-26
Add FAQ for GitHub for Windows

zstadler:


tpope:
Is this for an out of the box install of GitHub for Windows, and if so, is it still applicable in 2019?

zstadler:
It seems this is no longer needed. Thanks!
I've tested `vim-fugitive` on Windows without the above setup and it works out-of the box with "Github Dosktop -> Repository -> Open in Command Prompt".

PR #872, opened 2017-02-02
Stricter match for fugitive:// buffers

### LINKED ISSUE #871: Over-matching fugitive buffer breaks path for some buffers
The pattern match for fugitive buffers is against '//' which can match some (possibly mistaken but still valid) file paths. This causes a change to `&path` when it isn't expected.

Match was introduced https://github.com/tpope/vim-fugitive/commit/063d1fcaa980207e44f2bcbe1bccf895daf2b3af

Currently it's at https://github.com/tpope/vim-fugitive/blob/b754bc2031f21a532c083dd0d072ba373bbe3a37/plugin/fugitive.vim#L198

I think changing it to `fugitive://` should work

natebosch:
Fixes #871

It is valid - though odd - to open a file at `some//path`. In that case
the current check for fugitive buffers matches and changes &path
unexpectedly. A stricter match against `fugitive://` prevents this.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -195,7 +195,7 @@ function! fugitive#detect(path) abort
       nnoremap <buffer> <silent> y<C-G> :call setreg(v:register, <SID>recall())<CR>
     endif
     let buffer = fugitive#buffer()
-    if expand('%:p') =~# '//'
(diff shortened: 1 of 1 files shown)


natebosch:
I confirmed that fugitive buffers, for example `:Gdiff` still get the expected `&path`

PR #881, opened 2017-02-17
Add keybindings for interactive rebase

TheLocehiliosan:
Hi Tim,

I've been using these keybindings for about two years and they've worked well. I realize this is a bit similar to PR #334, but I think the issue you had with that change was the map delays. It shouldn't be a problem with these bindings.

p **I** ck
**R** eword
**E** dit
**S** quash
**F** ixup
e **X** ec
**D** rop

I think these bindings are simpler, easy to remember and don't collide with any *useful* mappings for most editing of interactive rebases. They are simply the single letter version of the commands, but uppercase. The only exception is for "pick". `P` was too useful for editing, and I thought about using `K`, but `keywordprg` is also useful for viewing commit details. Since "pick" is the default command, usually it isn't necessary to change anyway.

For reference, this is the output from `git rebase -i`:

```
# Commands:
# p, pick = use commit
# r, reword = use commit, but edit the commit message
# e, edit = use commit, but stop for amending
# s, squash = use commit, but meld into previous commit
# f, fixup = like "squash", but discard this commit's log message
# x, exec = run command (the rest of the line) using shell
# d, drop = remove commit
```
I hope you consider merging this. It makes ripping through a rebase very quick and easy.

Cheers

-Tim

### DIFF
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -306,6 +306,24 @@ C                       Go to the commit containing the current file.
 a                       Show the current tag, commit, or tree in an alternate
                         format.
 
+                                                *fugitive-interactive-rebase*
+These maps are available during an interactive rebase, and change the command
+selected for the current line.
+
+I                       pick: use commit
+
+R                       reword: use commit, but edit the commit message
+
+E                       edit: use commit, but stop for amending
+
+S                       squash: use commit, but meld into previous commit
+
+F                       fixup: "squash", but discard this commit's log message
+
+X                       exec = run command (the rest of the line) using shell
+
+D                       drop = remove commit
+
 SPECIFYING REVISIONS                            *fugitive-revision*
 
 Fugitive revisions are similar to Git revisions as defined in the "SPECIFYING
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -3097,3 +3097,17 @@ augroup fugitive_foldtext
         \    set foldtext=fugitive#foldtext() |
         \ endif
 augroup END
+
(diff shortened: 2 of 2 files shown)


TheLocehiliosan:
I updated this PR so the binding for `exec` actually creates a new line, leaving the user in insert mode to type the command to execute. I also added the bindings to the documentation.

tpope:
These 7 mappings shadow 7 built-ins that are perfectly reasonable to use in a plain old file like `rebase-todo`. It is perfectly reasonable, for example, to change a `pick` line with `Redit<Esc>`, and if a user in the habit of doing that suddenly got a `reword` instead, they would be justified in throwing this plugin in the garbage for breaking Vim. I will never allow something like this in Fugitive.

Fear not, though, the new `:Gstatus` has a section for the current rebase todo, and that's not a file you can edit, so it's perfect for this sort of stuff. It's been on my todo list for a while now to add maps like these. I don't think it's gonna make 3.0 but hoping to have it for 3.1.

If you'd like to help, what I first need is a function that takes a verb and an abbreviated commit SHA and updates that SHA to have the new action in `.git/rebase-merge/git-rebase-todo` using `readfile()` and `writefile()`. This is more involved than a `:substitute` but certainly achievable for anyone willing to get their hands dirty with some VimL.

The other half is hooking this function up to the `:Gstatus` buffer, which involves a lot of Fugitive internals and probably best left to me.

TheLocehiliosan:
> These 7 mappings shadow 7 built-ins...
> I will never allow something like this in Fugitive.

I understand this reasoning. Thanks for reviewing it. Cheers.

PR #889, opened 2017-03-20
Trigger Autocommands after changing windows from autocmd

chrisbra:
By default autocommands do not nest and if fugitive#reload_status() is
called from an autocommand vim might switch windows internally to
refresh its status.

Unfortunately, after moving back to the original window, it does not
trigger the WinEnter autocommand and since airline depends on it to
display the correct statusline, the statusline highlighting will be
wrong.

Fix this by explicitly calling WinEnter autocommands, if windows have
been changed.

closes vim-airline/vim-airline#1452

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -748,8 +748,8 @@ call s:command("-bar Gstatus :execute s:Status()")
 augroup fugitive_status
   autocmd!
   if !has('win32')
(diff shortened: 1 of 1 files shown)


tpope:
I just added `nested` to the autocommands to fix this. I feel like if we fire `WinEnter` but not `WinLeave`, this will create issues for a different set of plugins.

PR #893, opened 2017-04-05
Fix temp_files cache on Windows when TEMP is set to a short path

dummyunit:
If `%TEMP%` is set to a short path (e.g. "C:\LongDi~1") then `tempname()` will return a file name that contains that short path. If that path is later used as key for entry in `s:temp_files` dictionary, that entry won't be found in BufNewFile,BufReadPost events because `<afile>` is `expand()`'ed before it is used as a key for `s:temp_files`.
In the end, user gets cryptic error message about "C:\LongDirName\VI12345.tmp.fugitiveblame" not being a git repository when he tries to open a commit in Gblame window.

To workaround that we expand paths of temp files when adding entries to `s:temp_files`.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1390,6 +1390,9 @@ function! s:Edit(cmd,bang,...) abort
       return 'redraw|echo '.string(':!'.git.' '.args)
     else
       let temp = resolve(tempname())
+      if has('win32')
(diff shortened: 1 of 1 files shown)


tpope:
Could you remove the inline comments please? Anyone that wants an explanation can see your commit message with `:Gblame` :) :)

dummyunit:
I have updated the commit, moved relevant information from comments into commit message.

PR #902, opened 2017-04-28
Protect against symlinks.

### LINKED ISSUE #831: Empty buffer_path when repo tree is a symlink
When in /home/user/.dotfiles/vim/neobundles/ultisnips and with a symlink `/home/user/.vim -> .dotfiles/vim/`:

% vim /home/user/.vim/neobundles/ultisnips/pythonx/UltiSnips/snippet_manager.py

The buffer shows up as "pythonx/UltiSnips/snippet_manager.py", apparently since Vim resolves the symlink.

When calling `:Gdiff` now, I get weird errors like nothing happening on the first call, and then a dir manager opening up (netrw and/or Dirvish).  This is causes by `s:buffer_path` returning an empty string.

While I think there should be a safeguard in `buffer_path`, I could fix it using the following patch:

``` diff
diff --git i/plugin/fugitive.vim w/plugin/fugitive.vim
index 345ea34..a4be745 100644
--- i/plugin/fugitive.vim
+++ w/plugin/fugitive.vim
@@ -589,10 +589,17 @@ function! s:buffer_path(...) dict abort
   elseif s:cpath(self.spec()[0 : len(self.repo().dir())]) ==#
         \ s:cpath(self.repo().dir() . '/')
     let rev = '/.git'.self.spec()[strlen(self.repo().dir()) : -1]
-  elseif !self.repo().bare() &&
-        \ s:cpath(self.spec()[0 : len(self.repo().tree())]) ==#
+  elseif !self.repo().bare()
+    if s:cpath(self.spec()[0 : len(self.repo().tree())]) ==#
         \ s:cpath(self.repo().tree() . '/')
-    let rev = self.spec()[strlen(self.repo().tree()) : -1]
+      let rev = self.spec()[strlen(self.repo().tree()) : -1]
+    else
+      let resolved_tree = resolve(self.repo().tree())
+      if s:cpath(self.spec()[0 : len(resolved_tree)]) ==#
+        \ s:cpath(resolved_tree . '/')
+        let rev = self.spec()[strlen(resolved_tree) : -1]
+      endif
+    endif
   endif
   return s:sub(s:sub(rev,'.\zs/$',''),'^/',a:0 ? a:1 : '')
 endfunction
```

What do you think?  Should I create a PR for that?

mcepl:
Fix #831 

This is just the patch from that issue reformulated as the pull request. This seems to fix (or workaround?) for https://github.com/airblade/vim-rooter/issues/60

``b:git_dir`` is still unexpanded, but that probably doesn’t matter.

@tpope and @airblade , what do you think?

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -595,10 +595,17 @@ function! s:buffer_path(...) dict abort
   elseif s:cpath(self.spec()[0 : len(self.repo().dir())]) ==#
         \ s:cpath(self.repo().dir() . '/')
     let rev = '/.git'.self.spec()[strlen(self.repo().dir()) : -1]
-  elseif !self.repo().bare() &&
-        \ s:cpath(self.spec()[0 : len(self.repo().tree())]) ==#
+  elseif !self.repo().bare()
+    if s:cpath(self.spec()[0 : len(self.repo().tree())]) ==#
         \ s:cpath(self.repo().tree() . '/')
-    let rev = self.spec()[strlen(self.repo().tree()) : -1]
+      let rev = self.spec()[strlen(self.repo().tree()) : -1]
+    else
+      let resolved_tree = resolve(self.repo().tree())
+      if s:cpath(self.spec()[0 : len(resolved_tree)]) ==#
+        \ s:cpath(resolved_tree . '/')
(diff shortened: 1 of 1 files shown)


tpope:
I think I like my proposed solution in #831 better. :)

PR #904, opened 2017-05-01
Fix readme links to Gbrowse plugins

bronzehedwick:
Github changed the way they parse markdown, breaking a bunch of patterns.


tpope:
Thank you for the vote of confidence but this looks like my own screw up, not GitHub's.

PR #909, opened 2017-05-07
throw the joined errors if error is empty on commit

### LINKED ISSUE #907: Empty error message when commit fails with "useConfigOnly = true"
I have recently added the `useConfigOnly = true` option to my git configuration.
When trying to commit from `Gstatus` window with `cc` on a repo where my user properties are not defined, fugitive fails (which is expected) but the error message reported is empty: `fugitive: `.

Is there any way to tell fugitive to be more verbose in this particular use case?

gautaz:
This should fix #907.

Perhaps joining the whole errors array is a bit harsh but at least the user will have a full understanding of why git failed during the commit.

In the particular use case of the related bug, the message becomes:
`fugitive:  *** Please tell me who you are.  Run    git config --global user.email "you@example.com"   git config --global user.name "Your Name"  to set your account's default identity. Omit --global to set the identity only in this repository.  fatal: no name was given and auto-detection is disabled`

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1118,7 +1118,7 @@ function! s:Commit(args, ...) abort
       elseif error ==# '!'
         return s:Status()
       else
-        call s:throw(error)
+        call s:throw(error?error:join(errors, ' '))
       endif
     endif
   catch /^fugitive:/


tpope:
* That ternary isn't doing what you expect (you might want `!empty(error)`).
* Your commit message is not in line with the contribution guidelines.

PR #910, opened 2017-05-08
Throw joined errors if error is empty on commit

gautaz:
Fix tpope/vim-fugitive#907
Replace tpope/vim-fugitive#909

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1118,7 +1118,7 @@ function! s:Commit(args, ...) abort
       elseif error ==# '!'
         return s:Status()
       else
-        call s:throw(error)
+        call s:throw(empty(error)?join(errors, ' '):error)
       endif
     endif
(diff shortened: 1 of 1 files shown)


gautaz:
@tpope Does this PR complies better with your guidelines than #909?
If not, can you detail please?

gautaz:
@tpope I guess you are a little busy for now, can you please just confirm that the PR will be considered at some time ?

tpope:
I've been hesitant to take this approach because it can result in a *really* long single line error, but I guess I'll give it a try.

Chill with the aggressive followups, you're not helping your case.

gautaz:
Sorry, I did not mean to be rude, as I'm not an english native, I wasn't sure I had followed the guidelines in the right way.
Thanks for the update and for trying the PR, I'd be happy to modify it if an acceptable alternative is found.

PR #913, opened 2017-05-12
Support the master... format in :Gdiff

### LINKED ISSUE #876: Accept ':Gdiff master...'
First of all, thanks for the fantastic plugin !

While `:Gdiff master` works perfectly, `:Gdiff master...` (master... is supposed to give you the commit at where your current branch forked from master) returs some junk, like it opens a new window but with some random commit, with its log and all files instead of the current file only.

LemonBoy:
It is implemented asking git what is the common ancestor between the
specified revision and the HEAD.
Closes #876

### DIFF
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -316,6 +316,7 @@ and the work tree file for everything else.  Example revisions follow.
 Revision        Meaning ~
 HEAD            .git/HEAD
 master          .git/refs/heads/master
+commit...       The common ancestor between commit and HEAD
 HEAD^{}         The commit referenced by HEAD
 HEAD^           The parent of the commit referenced by HEAD
 HEAD:           The tree referenced by HEAD
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1827,6 +1827,13 @@ function! s:Diff(vert,keepfocus,...) abort
       catch /^fugitive:/
         return 'echoerr v:errmsg'
       endtry
+    elseif arg =~# '[^.]\.\.\.$'
+        let base = s:repo().git_chomp('merge-base',matchstr(arg, '\S*[^.]'),'HEAD')
+        try
+            let file = s:repo().rev_parse(base).s:buffer().path(':')
+        catch /^fugitive:/
+            return 'echoerr v:errmsg'
+        endtry
     else
       let file = s:buffer().expand(arg)
     endif


tpope:
This is really specific. Shouldn't we generalize to `commit1...commit2`?

LemonBoy:
since `Gdiff` does a diff against `HEAD` I didn't add support for specifying the second endpoint and `HEAD` is implicitly assumed, the first endpoint can be whatever string that `parse_rev` accepts.

LemonBoy:
I've reworded the help so that it is less ambiguous, do you need any other modifications to be done?

niklaas:
What is the status of this MR? This feature would be really helpful. Do you need any help?

tpope:
`Gdiff` *doesn't* do a diff against `HEAD` by default -- it diffs against either the work tree or index version -- and regardless I think the current file's commit, not `HEAD`, would be a better default. That is, if you've done `:Gedit my-branch:file`, then `:Gdiff other-branch...` would make more sense as `other-branch...my-branch` not `other-branch...HEAD`. But this isn't a use case I can recall having, personally, so maybe I'm missing something.

niklaas:
Hm, you're right, instead of proposing to implement a particular feature, let's talk about use cases here. 

Where I work, we use the "feature branch workflow" (https://www.atlassian.com/git/tutorials/comparing-workflows/feature-branch-workflow). So, in my daily work, I often checkout feature branches of others and want to see the changes they made to master *from the point when they were branching of their feature branch from master*.

To make this clear: It happens that a lot of others had already merged their feature branches into master. But, when looking at the branch I review, I am not interested in their changes; I am only interested in the changes that were made "within" (if one can say it that way) the feature branch I review. To achieve this, being on the feature branch, I can use `git diff master...` because it leverages `git merge-base master <feature-branch>`.

So, indeed, when I think about it now, doing something like `:Gdiff master...` doesn't really do justice to `:Gdiff` because it expects a single `fugitive-revision`. I have already tried things like `:Gdiff !git merge-base master HEAD` but that doesn't work because it seems `:Gdiff` interprets the argument literally.

Thus, `:Gdiff master...` isn't what we talk about here actually. (I can only speak for myself but I assume @LemonBoy thinks about the same use case.) Rather, we'd need another `fugitive-revision` that translates to `git merge-base <sha-1> <sha-2>`. Just thinking aloud here.

blueyed:
> I have already tried things like `:Gdiff !git merge-base master HEAD` but that doesn't work because it seems `:Gdiff` interprets the argument literally.

What about `:exe ':Gdiff '.system('git merge-base master HEAD')`?

Maybe a fugitive revisions for "merge-base" could be added, i.e. something like `:mb:master`?

tpope:
The lack of a native Git revision syntax for the merge base is the crux of an issue. How would you even do this on the command line? I think you're stuck with some abomination like `git diff $(git merge-base branch HEAD)`, which is why the Fugitive version is an equally hideous `:exe`.

I think repurposing the `...` notation (which is otherwise nonsensical in this context) is probably better than inventing a new one like `:mb:master` (which sort of overlaps with index files with colons in the name). Though I'm not really a fan of either.

uedan-slw:
Have not this pull request merged ?
I met same issue, Thanks.

tpope:
As an undocumented addition to `:Gdiffsplit` I would be fine with merging an updated version of this. It needs to use `:sTreeChomp()` and `fugitive#RevParse()` now rather than the old `s:repo()` functions.

teto:
hourra \o/ You rock o/

tpope:
`git checkout a...b` treats the range as a merge base. I think it's fine to make this official syntax.

PR #935, opened 2017-08-17
Removing change directory step from pathogen installation step.

Khalian:


tpope:
Breaks the very next line.

Khalian:
@tpope Are you referencing vim -u NONE -c "helptags vim-fugitive/doc" -c q?

There should be no reason a developer needs to be under a certain directory to launch vim.

tpope:
`:helptags` with a relative path most certainly depends on the current working directory.

Khalian:
Ok, does the path change work on the vim arg good?

tpope:
New rule: no bikeshedding installation instructions. I maintain them for 50ish Vim plugins and they're already badly out of sync without external meddling.

PR #936, opened 2017-08-27
Vim8 terminal support

tracyone:
see following video:

https://asciinema.org/a/IRbMiGITFNl3I5SMoo06DxQOI

update video

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -724,15 +724,19 @@ function! s:Git(bang, args) abort
     let git .= ' --no-pager'
   endif
   let args = matchstr(a:args,'\v\C.{-}%($|\\@<!%(\\\\)*\|)@=')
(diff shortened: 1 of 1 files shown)


coot:
This indeed fixes the fugitive in vim with `+terminal` feature. @tracyone could you remove the dot to fix the commit message.

@tpope sooner or later any fugitive user will be hit by this.  vim devs are working hard on the `+terminal` feature. Not everything works yet (from the more important things scrolling is not working, but beside that the vim terminal works pretty well), and there is steady progress on it.

tracyone:
@coot In this patch, I split window instead of a new tab.Now I think maybe a new tab is a better choice. What do you think?

tracyone:
Done. Same action with neovim.

coot:
@tracyone why did you closed this? Was it fixed in fugitive?

PR #538, opened 2014-08-22
fix :diffoff!

coot:
`:diffsplit` records settings (like `cursorbind`, `scrollbind`, ...), and
`:diffoff` resotres them.  So if you want to set up `cursorbind` it is
better after executing `:diffsplit`.

PR #937, opened 2017-08-27
Fix diffoff command.

coot:
`:diffsplit` records settings (like `cursorbind`, `scrollbind`), and
`:diffoff` resotres them.  So if you want to set up `cursorbind` it is
better after executing `:diffsplit`.

Version of #538 but rebased on master.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1853,9 +1853,6 @@ function! s:Diff(vert,keepfocus,...) abort
     let spec = s:repo().translate(file)
     let commit = matchstr(spec,'\C[^:/]//\zs\x\+')
     let restore = s:diff_restore()
-    if exists('+cursorbind')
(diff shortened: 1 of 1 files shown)


coot:
I am not really sure why the git commit message check fails.

sinisterstuf:
It could be the full stop at the end of the summary line.

coot:
Indeed it was, the full stop.

tpope:
Oh, the fact Vim restores rather than resets to default was the piece I was missing. I don't really like bouncing back and forth between windows like that - I assume it fires a bunch of unnecessary `WinEnter` and `WinLeave` events - but if the current behavior is causing a problem I'll accept it.

tpope:
I've removed the line because Vim automatically sets `'cursorbind'` for diff buffers. I assume I added it because there was a window after the feature was added before it was set automatically.

PR #945, opened 2017-09-27
Documented how to get to the current file from any fugitive-revison

srasku:


lingz:
+1

tpope:
This is just `:Gedit %` now. Resubmit if you want, but there's a lot of `%` examples so hopefully it goes without saying? (It's also the default, so probably not very useful to most people.)

Issue #944, opened 2017-09-21
Feature request: Checkout multiple files at once using visual mode


aankittcoolest:
It would be cool to checkout multiple files at once by using virtual selection of vim.

dhruvasagar:
@aankittcoolest Kindly check out my fork and see if you face any issues while tpope finds the time to merge the PR

tpope:
@dhruvasagar, what happens to the multiple `echomsg` lines? The reason I haven't hooked this up myself is I want the recovery situation to be top notch before allowing the user to shoot themselves in the foot this hard.

PR #948, opened 2017-10-06
Checkout multiple files in visual mode #944

dhruvasagar:
Changes to allow checkout of multiple files in visual mode from the Git Status window, similar in function to how Staging works. Let me know what you think.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -880,28 +880,40 @@ function! s:StageReloadSeek(target,lnum1,lnum2) abort
   call search('^#\t\%([[:alpha:] ]\+: *\|.*\%uff1a *\)\=\V'.jump.'\%( ([^()[:digit:]]\+)\)\=\$','W')
 endfunction
 
-function! s:StageUndo() abort
-  let [filename, section] = s:stage_info(line('.'))
-  if empty(filename)
-    return ''
-  endif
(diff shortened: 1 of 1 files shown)


dhruvasagar:
Can someone take a look at this ? Would love if this can be merged.

dhruvasagar:
@tpope any thoughts ?

tpope:
@dhruvasagar, what happens to the multiple echomsg lines? The reason I haven't hooked this up myself is I want the recovery situation to be top notch before allowing the user to shoot themselves in the foot this hard.

(Reposting from #944 because it really belongs here.)

dhruvasagar:
@tpope Yea even I haven't been using this particularly, although I do think it may be useful. I can remove the echomsg, are there any other changes you think I should make ? (i'll rebase with latest)

tpope:
My *entire* objection to this feature is that I want to make sure it is possible to undo if you accidentally checkout multiple files. Deleting the `echomsg` would hide the undo instructions and literally make it impossible to undo. So no, do not remove it; it's the worst possible thing you could do.

I would not bother rebasing at this time. Adding the while loop is easy enough. It's figuring out how to handle undo that's hard.

dhruvasagar:
Yea this one was almost 2 years old, looks like a lot has changed. Thanks.

PR #949, opened 2017-10-09
Do not accept :Gread count in argument position

mbudde:
Use `-range=0` instead of `-count` to prevent Vim from interpreting a
branch name starting with a number as a count, e.g. `:Gread 123-foo`
being interpreted as `:123Gread -foo`.


tpope:
Can you change this to `-range=-1`, along with updating the ternary to `<count> == -1 ? '' : <count>`? Years of dealing with bugs around various `-range` variants have led me to this as the least quirky approach. Thanks!

mbudde:
I have updated the commit to use `-range=-1` instead :+1:

PR #954, opened 2017-10-24
Added git checkout --theirs/--ours support for Gstatus

vicentebolea:
When merging conflicting files using `git checkout --theirs/--ours file && git add file` is a very popular command, so much that some people we have an alias for this operation. Why not having it in the _fugitive_ `:Gstatus` window. 

If we include this feature, many of us does not have to leave VIM to perform this operation, we can easily just press one key stroke in each of the files.

Note that I am not a experienced _vimscript_ developer. My small hackish function just works in my configuration, it might not work in other configurations.

I updated the docs as well.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -1567,6 +1567,8 @@ function! fugitive#BufReadStatus() abort
     nnoremap <buffer> <silent> r :<C-U>exe <SID>ReloadStatus()<CR>
     nnoremap <buffer> <silent> R :<C-U>exe <SID>ReloadStatus()<CR>
     nnoremap <buffer> <silent> U :<C-U>echoerr 'Changed to X'<CR>
+    nnoremap <buffer> <silent> st :<C-U>execute <SID>StageTheirs()<CR>
+    nnoremap <buffer> <silent> so :<C-U>execute <SID>StageOurs()<CR>
     nnoremap <buffer> <silent> g<Bar> :<C-U>execute <SID>StageDelete(line('.'),v:count)<CR>
     xnoremap <buffer> <silent> g<Bar> :<C-U>execute <SID>StageDelete(line("'<"),line("'>")-line("'<")+1)<CR>
     nnoremap <buffer> <silent> X :<C-U>execute <SID>StageDelete(line('.'),v:count)<CR>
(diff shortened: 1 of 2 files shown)


davclark:
This would be clutch - I'm in `git checkout --theirs` hell as we speak. Is there anything I can do to get this merged? Is there any problem with the implementation, or just a lack of rebasing on current master?

vicentebolea:
@davclark I almost forgot about this PR. I updated to the latest version of this project. Let's hope its accepted!

davclark:
This was my first attempt... I just fetched / checked out this pr like so:

```
git fetch origin pull/954/head:ourstheirs
git checkout ourstheirs
```
The docs updated correctly, and the `st` command was available from a `:Gstatus` buffer, but when executed, I got the following (I'm running this in Oni, which is a UI in front of neovim, and it won't let me copy-paste text, hence the screenshot):

screenshot from 2019-01-21 14-08-04

Looks like somewhere, `s:stage_info()` got rewritten as `s:StageInfo()`.

davclark:
Works! Even works in visual mode to stage multiple files at once.

@tpope I hope you are willing to entertain accepting this PR!

vicentebolea:
@tpope, before merging let me rebase the changes to the tip of `master` and write a consistent good commit message for this feature.

vicentebolea:
@davclark Thanks for the feedback

vicentebolea:
@davclark @tpope ready to go!

vicentebolea:
@tpope I started this a year ago when I found myself with the need to easily pick a version (ours or their) within `:Gstatus` during a large merge. Since then I have being using this branch of mine and it has been a great way to speedup my merges. Also, apparently some other people has found this useful such as @davclark.

Currently, the alternative would be to go to a terminal and type `git checkout --ours` for each of the files and later refresh `:Gstatus`. 

As for the confusion of getting mix up _theirs_ and _ours_, I agree that those terms are confusing at times, however, it seems to be what `git` refers to the `current ` or `to-merge` version of the file.

Any idea to make this easier to use or to be in line with the `vim-fugitive` project, I will be happy to code it up :)

tpope:
What I mean is I must have resolved a thousand merge conflicts by this point in my life and I can count on one hand the number where the answer was "discard the changes one branch", and even in those cases, it's not apparent until I've already started manually merging. What sort development workflow do you have that leaves you regularly, confidently, blindly saying "just use my branch, I'm sure it's fine".

The one time (there was only one time) I saw someone else do it, they were throwing away valid changes by mistake. The **last** thing I want to do is provide a feature that endorses such recklessness.

vicentebolea:
__TL;DR__ To sum up, this PR adds a shortcut for a `git` common feature which aids to complete one of the possible scenarios of merging two versions of a file, which is to choose one of the versions.

> What sort development workflow do you have that leaves you regularly, confidently, blindly saying "just use my branch, I'm sure it's fine".

In a fork model some of the `metadata` files such as README, `Makefiles`, `configure.ac` have to be kept at times on one of the version (Not a merge of two). This is the use case that I have encountered several times in my work which demanded me to update to an upstream new version my current fork, in that case there were a subset of files that I could confidently `git checkout --ours files..` them. 

The might be other uses, we could use the same rationale of why git decided to provide `checkout --ours/--theirs`

> The one time (there was only one time) I saw someone else do it, they were throwing away valid changes by mistake. The last thing I want to do is provide a feature that endorses such recklessness.

You are talking from one experience, we should not make a rule on that. This feature gives a power which is up to the user to use adequately, by the same argument `vim` shall not provide `:q!` since I have seen somebody losing his work trying to exit the editor.

Still I understand your fear to provide an ease way to screw things up, but let me remind you that the user is always a `git checkout -m -- file` or `git merge/rebase --abort` away to recover the previous changes


> P.S. The confusing part is that it's backwards for rebasing. Well, not backwards. Sideways.

I could not agree more on this, when it comes to solve unmerge paths on a rebase, the `ours/theirs` is sort of reversed which is truly confusing.

davclark:
The specific example where this came up for me is on a relatively large code-base using a mono-repo for relatively independent aspects of a web application. I had a long-running branch in one "backend" package in the mono-repo, and in the meantime, some gnarly merge-hacking happened on the UI side. As the maintainer of my long-running branch, I knew that only one file in the whole UI codebase was modified - so the merge conflicts were just reproducing a mess someone else had created (and already solved).

In the end, I dropped into IPython and ran `git checkout --theirs <file>; git add <file>` via system calls over a list of the ui files. This was annoying, and also more error prone than a visual TUI.

tpope:
It has occurred to me that the (admittedly confusing) way we show merge conflicts in the staged and unstaged sections makes this an extremely natural addition to `X`. I'm also adding `2X` and `3X` as undocumented ways to force ours and theirs, if you refuse to think about which section is which.

Long term, I expect `X` to be replaced by an `x` that prompts for confirmation. Prompting for ours or theirs would then be a natural next step.

davclark:
Thanks for your thoughts on this! And please feel free to @mention me if you want review on a PR for that. I'm not a vimscript guy, but I can definitely read it a bit and certainly, I can test the behavior.

Issue #860, opened 2016-12-17
Minor typo in README.md


manasthakur:
`Git!` should be `:Git!`; just for consistency!

PR #958, opened 2017-10-26
Fix inconsistent documentation in README

asheidan:
Previously it listed `Git!` without leading colon.

This closes tpope/vim-fugitive#860


tpope:
Whoops, found this just a week or so after removing the line entirely. Thanks anyways.

PR #960, opened 2017-10-30
Allow custom fugitive#statusline prefix and suffix

ericbn:
fugitive#statusline('[',']')

outputs, for example

    [:7c9b87a(master)]

The suffix parameter is optional. If no parameters are given, the
original logic is applied.


tpope:
`[:` is just hideous. If you really must do this to yourself, use `substitute()`.

Issue #866, opened 2017-01-09
Gdiff to a revision without this file


somini:
If you run `Gdiff $revision` and that revision doesn't have that file (it was created afterwards), there's an error and Fugitive gets confused. If you repeat the command, the split gets opened with an empty file, and when you close that all returns to normal. Here's a test case.

- Open `$REDACTED_FILENAME` that doesn't exist yet on commit e51d2e933cf35a2d06174c6f667a345732e167b1.

```
Gdiff e51d2e933cf35a2d06174c6f667a345732e167b1
```

```
Error detected while processing BufReadCmd Auto commands for "fugitive://**//[0-9a-f][0-9a-f]*":
fugitive: rev-parse e51d2e933cf35a2d06174c6f667a345732e167b1:$REDACTED_FILENAME: fatal: Needed a single revision 
```

blueyed:
It depends on if you have `set hidden` (which most people will have): with `hidden` `:ls!` lists the fugitive buffer with `x` ("a buffer with read errors"), and when you issue the command again it appears to get used.
With `:set nohidden` however, you will get the same error again.

Related: #607

somini:
I can confirm `96adffd` solves this by diffing with an empty file.

PR #961, opened 2017-10-30
s:BufReadObject: catch rev-parse errors and assume empty object

blueyed:
With `:Gdiff ^` on a file that has just been added, you would get an
error via `git rev-parse --verify`:

> fatal: Needed a single revision

The error message is clearer when not using `--verify`:

> fatal: Path 'X' exists on disk, but not in 'SHA'.

The behavior depends on if you use `:set hidden`, where the second
invocation works, as in will open an empty buffer - apparently since it
has been created as "a buffer with read errors" ("x" in `:ls!`) before.

Fixes https://github.com/tpope/vim-fugitive/issues/866.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -2748,6 +2748,8 @@ function! s:BufReadObject() abort
     endtry
 
     return ''
+  catch /^fugitive: rev-parse/
+    return ''
   catch /^fugitive:/
     return 'echoerr v:errmsg'
   endtry


plugin/fugitive.vim
```diff
@@ -2748,6 +2748,8 @@ function! s:BufReadObject() abort
     endtry
 
     return ''
+  catch /^fugitive: rev-parse/
+    return ''
```
blueyed: Could have an `:echom` with the error maybe?

somini:
This fixes #866

PR #963, opened 2017-11-03
Add Gbrowse returning shortened URL

mcepl:
When communicating over IRC with somebody and I want to point him to
some particular piece of code, it would be useful if Gbrowse added to
the clipboard **shortened** URL of the code in question.

This PR adds such functionality. It is switched off per default and it
could be enabled by settings global variable
``g:fugitive_short_shorten`` to 1.

In case the extension is used inside of the corporate network and
internal addresses should not be leaked to the public URL shorteners,
one could set up dictionary ``g:fugitive_short_private_domains`` with
key being string which matches all domains inside of the private
network, and value is URL of the internal URL shortener. If this
dictionary is not set, the script uses ``https://da.gd/s?url=`` as
default.

For example, inside of the Red Hat network one can have in her
``~/.vimrc``

```viml
let g:fugitive_short_shorten = 1
let g:fugitive_short_private_domains = {}
let g:fugitive_short_private_domains['redhat.com'] =
    \ 'https://url.corp.redhat.com/new?'
```

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -90,7 +90,7 @@ directory.  Edit a file from the repository.
 
 `:Gbrowse` delegates to `git web--browse`, which is less than perfect
 when it comes to finding the right browser.  You can tell it the correct
-browser to use with `git config --global web.browser ...`.  On macOS, for
+browser to use with `git config --global web.browser ...`.  On OS X, for
 example, you might want to set this to `open`.  See `git web--browse --help`
 for details.
 
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -12,6 +12,10 @@ if !exists('g:fugitive_git_executable')
   let g:fugitive_git_executable = 'git'
 endif
 
+if !exists('g:fugitive_short_shorten')
+    let g:fugitive_short_shorten = 0
+endif
+
 " Section: Utility
 
 function! s:function(name) abort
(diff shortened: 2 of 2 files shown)


mcepl:
OK, I am sorry, I don't know what's wrong with the commit message.

justinmk:
There's no reason for this to be baked-into fugitive or any other plugin. It could be achieved by:

- a mapping
- an IRC bot ...
- /etc/hosts or some other internal policy

shumphrey:
I assume it fails because:
>  Write your commit message in the imperative: "Fix bug" and not "Fixed bug"
or "Fixes bug."  This convention matches up with commit messages generated
by commands like git merge and git revert.

mcepl:
> There's no reason for this to be baked-into fugitive or any other plugin. It could be achieved by:
> 
>     a mapping
>     an IRC bot ...
>     /etc/hosts or some other internal policy

I am afraid, I don't understand.

1. Could you suggest how to make mapping which would convert output from Gbrowse without any plugin?

2. How is IRC bot saving work against copying from clipboard to the Firefox? (not mentioning that I don't want to install random IRC bots on various IRC networks I am logged to).

3. Completely lost. How do I convert with ``/etc/hosts`` URL from https://gitlab.com/m2crypto/m2crypto/blob/4865bec83c1d7eabe9a574a76a54ad1c2987bcfb/M2Crypto/Rand.py#L19-36 to https://da.gd/2J5k ?

mcepl:
OK, so hopefully finally I have persuaded GitHub I have lovely commit message (I don't think so, but it doesn't matter). @tpope, could you take a look, please?

mcepl:
OK, API for ``Gbrowse`` plugins doesn't seem to be described anywhere, but it is certainly different from what this PR expects. Oh well.

PR #965, opened 2017-11-14
Fix Gblame not being able to show any specific commit on Windows

filcab:
This is adding the same fix that 4f24757df2a57fc01900334114b1790fcb75ad54 did,
but this time to ReplaceCmd.

The problem was that expand() can't expand short paths if the file doesn't
exist. So we split it in head and tail, and expand the head.


tpope:
This is encapsulated in a helper function used everywhere now.

PR #971, opened 2017-11-20
Use NERDTree.root instead of deprecated NERDTreeRoot

ledhed2222:
Looks like NERDTree depcrecated NERDTreeRoot and briefly removed it. They added it back as a copy of NERDTree.root (the correct variable) with a note that it should be removed in later versions. Let's use the correct variable in case they remove it again

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -241,7 +241,12 @@ augroup fugitive
   autocmd!
   autocmd BufNewFile,BufReadPost * call fugitive#detect(expand('%:p'))
   autocmd FileType           netrw call fugitive#detect(expand('%:p'))
-  autocmd User NERDTreeInit,NERDTreeNewRoot call fugitive#detect(b:NERDTreeRoot.path.str())
+  if exists("b:NERDTree.root")
+    autocmd User NERDTreeInit,NERDTreeNewRoot call fugitive#detect(b:NERDTree.root.path.str())
+  elseif exists("b:NERDTReeRoot")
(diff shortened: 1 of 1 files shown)


agibralter:
Would it be possible to move this forward?

plugin/fugitive.vim
```diff
@@ -241,7 +241,12 @@ augroup fugitive
   autocmd!
   autocmd BufNewFile,BufReadPost * call fugitive#detect(expand('%:p'))
   autocmd FileType           netrw call fugitive#detect(expand('%:p'))
-  autocmd User NERDTreeInit,NERDTreeNewRoot call fugitive#detect(b:NERDTreeRoot.path.str())
+  if exists("b:NERDTree.root")
+    autocmd User NERDTreeInit,NERDTreeNewRoot call fugitive#detect(b:NERDTree.root.path.str())
+  elseif exists("b:NERDTReeRoot")
```
agibralter: Shouldn't this be `b:NERDTreeRoot`?

ledhed2222:
@agibralter it's already supported where? by vim-fugitive?

agibralter:
@ledhed2222 I think you need to rebase. Check out master: https://github.com/tpope/vim-fugitive/blob/f3ccb0c12ee4985b8808f83059830a24cc92821c/plugin/fugitive.vim#L244

tpope:
Resolve the conflict and I'll merge.

agibralter:
Heh, if he resolves the conflict, I think he'll realize that it's fixed already 😄

Issue #167, opened 2012-03-12
Can't initiate :Gstatus from (shell) command-ine


xenomachina:
Here's what I'd like to be able to do (from my shell):

```
vim -c Gstatus
```

When I try this, vim says:

```
Error detected while processing command line:
E492: Not an editor command: Gstatus
```

I also tried using a VimEnter autocmd to see if I could delay the command to a point when vim might be more "ready" to run Gstatus:

```
vim -c 'au VimEnter * Gstatus'
```

This actually does get me farther: now there's a split, and .git/index is opened in it, but I'm seeing the raw binary index, not the human readable output of "git status". I'm guessing some of fugitive's autocmds haven't been set up yet.

I also tried the same but replacing VimEnter with BufEnter and WinEnter, but those resulted in nothing at all happening. (and if they did work I'd need to add additional code to disable the autocmds after the initial firing, as I don't want Gstatus running every time I enter a buffer or window...)

PR #977, opened 2017-12-05
Support `vim -c Gcommand`

pavoljuhas:
Make fugitive commands available in the "-c cmd" arguments.
Use explicit check for vim startup instead of VimEnter event,
which is triggered only after executing "-c cmd".

This enables `vim -c "Ggrep pattern"` and also
resolves https://github.com/tpope/vim-fugitive/issues/167.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -247,7 +247,6 @@ augroup fugitive
         \ if exists('b:NERDTree.root.path.str') |
         \   call FugitiveDetect(b:NERDTree.root.path.str()) |
         \ endif
-  autocmd VimEnter * if empty(expand('<amatch>'))|call FugitiveDetect(getcwd())|endif
   autocmd CmdWinEnter * call FugitiveDetect(expand('#:p'))
 
   autocmd FileType git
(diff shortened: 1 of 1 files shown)


pavoljuhas:
@tpope - have you had a chance to look at this PR?

Besides making the `vim -c Gcommand` work, the only difference I can see from the current behavior happens for

```sh
$ cd /to/repo1
$ vim -c 'cd /to/repo2'
```

I am not sure if such use is likely enough to care about, but if so, it could be fixed by restoring the VimEnter autocommand.

tpope:
The issue is that firing `doautocmd Fugitive` before Vim has loaded means an arbitrary set of plugins are left out. I know that's a really narrow edge case, but I refuse to engage in making a janky feature even jankier. After 3.0, I may explore making the commands global. Until then, use `vim . -c 'Ggrep'`.

pavoljuhas:
> ... Until then, use `vim . -c 'Ggrep'`.

For some reason this does not work for a directory:
```sh
$ vim . -c "Ggrep fugitive"
"." is a directory
Error detected while processing command line:
E492: Not an editor command: Ggrep fugitive
```
Supplying file argument (even non-existing) instead of `.` works:
```sh
$ vim x -c "Ggrep fugitive"
... 633 matches ...
```

tpope:
Netrw is the gift that keeps on giving 🙄🙄🙄

pavoljuhas:
`vim -c "Ggrep pattern"` works as of ec918b9a79512e82e00901d93bfb6d5f57bf56f0.  🎉

PR #984, opened 2017-12-29
Implement pushing the current branch upstream

parvus-mortalis:
- Copy fugitive#head to Gpush and Gfetch section to use for branch name
- Add Gpushup command for pushing the current branch to origin remote

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1657,8 +1657,17 @@ augroup END
 
 " Section: Gpush, Gfetch
 
+function! s:Head(...) abort
+  if !exists('b:git_dir')
+    return ''
+  endif
+
+  return s:repo().head(a:0 ? a:1 : 0)
+endfunction
+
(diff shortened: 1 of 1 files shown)


h3xx:
-1 for hard-coding the remote name `origin`.

The first thing I do when I clone a repo is `remote rename origin o` to save typing. There's no guarantee that *any* git repository has a remote, much less one names `origin`.

rsrchboy:
FWIW, when I fork I often make my remote `origin`, and name where I forked from `upstream`, so... yeah.  TMTOWTDI :)

tpope:
A good rule of thumb is that if your new Vim command (`:Gpushup`) would make just as much sense as new Git command (`git pushup`), it does not belong in Fugitive. If you want it for yourself, add it to your vimrc:

    autocmd User Fugitive command! -bar -bang Gpushup Gpush -u origin HEAD:master

Issue #369, opened 2013-08-06
Could we get some eye candy from GitHub blame in :Gblame?


bananu7:
I've recently discovered how nice GH blame really is. What caught my eye was:
- grouping lines affected by one commit
- coloring based on the date of change
- displaying a commit message that impacted the lines

I think at least the grouping should be relatively easy to implement, and it would significantly increase readability.

Something like that just looks like a wall of text:

![lounge](https://f.cloud.github.com/assets/2399997/917079/155a3be8-fe9d-11e2-9cd0-0c4c9f36546b.png)

compared to:
![lounge2](https://f.cloud.github.com/assets/2399997/917095/4d6236bc-fe9d-11e2-9836-145a47290cc0.png)

tpope:
Per commit colors are now standard and can be forced with `:Gblame --color-lines`. This is my solution. I will consider a @mattboehm style solution if it can be implemented with zero regressions ("breaks some buffer local keymaps" is a dealbreaker) and a performance overhead of less than double.

PR #986, opened 2018-01-04
Fix laggy blame scrolling

h3xx:
Don't add syntax groups for the blame "eye candy" from #369 unless we have the plugin to support it.

This speeds up highlighting if you don't have the plugin, since it doesn't make a group for each commit ID in the blame view.

When `&cursorline` was set, this feature caused CPU usage to max out while scrolling via spamming j/k.

In lieu of optimizing the syntax highlighting definition, or even turning it off with yet-another-config-variable, just omit adding the highlighting if adding them would effectively just create redundant highlight groups.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -2202,27 +2202,29 @@ function! s:BlameSyntax() abort
   hi def link FugitiveblameShort              FugitiveblameDelimiter
   hi def link FugitiveblameDelimiter          Delimiter
   hi def link FugitiveblameNotCommittedYet    Comment
-  if (&t_Co > 16 || has('gui_running')) && get(g:, 'CSApprox_loaded') && !empty(findfile('autoload/csapprox/per_component.vim', escape(&rtp, ' ')))
-    let seen = {}
-    for lnum in range(1, line('$'))
-      let hash = matchstr(getline(lnum), '^\^\=\zs\x\{6\}')
-      if hash ==# '' || hash ==# '000000' || has_key(seen, hash)
-        continue
-      endif
-      let seen[hash] = 1
-      if empty(get(s:hash_colors, hash))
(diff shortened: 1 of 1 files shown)


h3xx:
Note: This was always laggy, but it became unbearably so between vim 8.0.1175 and 8.0.1415, but I don't have time to fix that.

h3xx:
I've been using this change for a while, and it works well. I'd love to see some traction on this.

tpope:
CSApprox is needed to set `ctermfg` but it is *not* needed to set `guifg`. So, this breaks things for GUI and `'termguicolors'` users.

What is the upstream Vim issue, exactly? http://ftp.vim.org/pub/vim/patches/8.0/8.0.1175 looks pretty innocuous but also very clearly color related.

jeffvandyke:
FWIW, `:Gblame` scrolling is also slow for large files in Neovim v0.2.2

h3xx:
Okay I think I fixed it.

With [csapprox](https://github.com/godlygeek/csapprox) installed:

CLI

GUI

Without csapprox:

CLI

GUI

tpope:
It should work *without* CSApprox installed in the GUI.

tpope:
Tbh just wrap the whole thing in a `if get(g:, 'fugitive_colors', 1)` and be done with it.

h3xx:
Regretting this feature?

Anyway, I changed the name of the setting variable to `fugitive_blame_colors` to avoid collisions between any future settings, and to make it clear what it does. See 1295fd3a2b4e83866b9582236325ef2f3f2362ce

tpope:
I went vague because `fugitive_blame_colors` sounds like it might disable all highlighting in blame windows entirely. In the unlikely event I dream up a second ridiculous use of color, it will likely be just as gaudy and divisive, and I'd rather have one option rather than many if I can get away with it. Call it `fugitive_dynamic_colors` if you want to be descriptive, I guess.

h3xx:
Updated.

Issue #990, opened 2018-02-12
Gsdiff :1 opens vertical diff when diffopt contains vertical


HarmtH:
When ```diffopt``` contains ```vertical```, running ```Gsdiff``` with an argument (without argument works fine), the diff opens in a vertical split instead of a horizontal one.

tandrewnichols:
Not for me. Even without an argument `:Gsdiff` opens vertically.

tpope:
Can confirm. I think the solution would be to temporarily remove `vertical` from `diffopt` before splitting.

PR #991, opened 2018-02-12
Fix Gsdiff not always opening horizontal split

HarmtH:
Fixes issue #990 

```diffsplit``` obeys ```diffopt=vertical```, therefore use ```split``` and ```diffthis```, just like in the case where ```Gsdiff``` was called without arguments.


tpope:
Unfortunately fixed this before I saw this patch, but thanks!

Issue #600, opened 2015-02-02
:Gstatus mapping for git checkout --patch


sharplet:
In addition to interactively staging and unstaging files, I like to be able to interactively discard changes in the working directory with `git checkout -p`. It would be lovely to have a mapping for this in the `:Gstatus` buffer to go along with the `p` mapping!

I imagine the mapping would have the same effect on both staged and unstaged files, as in either case you're applying changes from the index to the working directory.

Thanks for the great plugin! :sparkles:

sharplet:
Actually come to think of it, it would make sense to have a non-interactive version as well! Just plain `:Git checkout <file>` to discard changes. This would go well with the `-` mapping.

tpope:
The non-interactive form is available (undocumented) on `U`. I guess `u` could be the interactive form. (`U` is a terrible mnemonic, but I want to keep it far away from the `c` family of commit mappings.)

sharplet:
u for unstage, right? Works for me ¯\_(ツ)_/¯

thirtythreeforty:
@tpope do you have plans to implement this?  It would be great to have!

tpope:
No plans, but I would entertain a pull request. I don't think it would be that hard to implement?

thirtythreeforty:
It probably would not be.  If I have some time perhaps I'll look at it!

Holzhaus:
I'd rather not use `u` for `git checkout -p` when `U` is the regular `git checkout`. In case you accidently have capslock on and you hit `u`, you lose all your worktree changes of that file. Hitting `U` is dangerous enough already. Why replace the `U` mapping with the patch version? In case you want discard the whole file you can still press `U` and then `a`.

tpope:
`X` (the replacement for `U`) now operates at a diff level, providing the same functionality as `checkout -p`.

PR #992, opened 2018-02-14
Add 'git checkout --patch' to Gstatus.

michaelainsworth:
Hi Tim. Thanks for your work. Love vim-fugitive.

I've added a mapping to support `git checkout --patch` functionality in `Gstatus`. This is in response to issue #600. I'd appreciate if you could review this pull request. If there's anything not up to scratch (commit log, code, etc), let me know and I'll try to address the issue.

The original commit log is below.

---

This commit adds the 'u' mapping to the buffer opened by Gstatus.  The
'u' mapping calls 's:UnstagedCheckoutPatch', which (by way of the
typical vim-fugitive mechanism) calls 'git checkout --patch <filename>'.

The 'u' mapping only works for single files (i.e., in normal mode). It
does not work for multiple files (e.g., in visual mode).

There are two 'safety mechanisms' at work when checking out a file using
the patch method. Firstly, the buffer is not automatically reloaded.
This allows you to overwrite (':w!') the file, should you accidentally
checkout a chunk. Secondly, a 'To restore...' message (similar to that
in 's:StageUndo') is displayed, allowing you to revert via the 'Git
cat-file blob' command.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -906,6 +906,26 @@ function! s:StageUndo() abort
   endif
 endfunction
 
+function! s:UnstagedCheckoutPatch() abort
+  let [filename, section] = s:stage_info(line('.'))
+  if empty(filename)
+    return ''
+  endif
+  let repo = s:repo()
+  let hash = repo.git_chomp('hash-object', '-w', filename)
+  if !empty(hash)
+    if section ==# 'unstaged'
+      let res  = 'Git checkout --patch -- ' . s:shellesc(filename)
+      let res .= '|checktime|redraw|echomsg ' .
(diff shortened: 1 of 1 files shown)


michaelainsworth:
Note that the status line of the first commit includes a trailing punctuation mark, failing the commit message criteria. I've added an empty commit with this punctuation mark removed. Apologies.

michaelainsworth:
(I just re-read CONTRIBUTING.markdown and noticed you want contributors to squash rebase then force push themselves. I've just done so. Again, apologies.)

tpope:
Looks good. Unfortunately I think the new hunk based checkout renders it obsolete.

Issue #685, opened 2015-09-03
:Ggrep with submodules


phmarek:
I've got a few checkouts that use submodules. Now `:Ggrep` has the problem that only the git checkout of the current file is searched, instead of all the git checkouts.

Perhaps there should be a test whether the current file is "just" in a submodule, and if it is, the git "root" should be used for the "git grep" and an additional "git submodule foreach --recursive grep" should be run (but see http://blog.endpoint.com/2010/04/make-git-grep-recurse-into-submodules.html).

tpope:
This sounds like a problem that I don't want to take ownership of.

jeaye:
I would love to have a `:Grgrep`, for recursive grep, within vim-fugitive. @tpope, are you opposed to having this new grep (not changing any existing greps)?

tpope:
So it would grep submodules but *not* the main repo? That sounds ... odd.

jeaye:
On Fri, Mar 18, 2016 at 11:22:11AM -0700, Tim Pope wrote:
> So it would grep submodules but *not* the main repo? That sounds ... odd.

You're right; it should grep both and concatenate the results.

tpope:
Show me a proof of concept I guess, because my hunch is there's at least half a dozen problems doing this you haven't even considered. That `"git grep ...; true"` hack isn't portable for example.

I think a more reasonable goal would be making it easier to compose your own command, perhaps by exposing the rewrite logic that happens afterwards.  Then whatever good-enough hack you come up with could leverage that.

odnoletkov:
git grep now supports --recurse-submodules (as do many other commands). So :Ggrep --recurse-submodules works as expected and there is even autocompletion for this option. There is also a git config option for that. So I guess this is resolved.

PR #999, opened 2018-03-14
Add option to customize 'git grep' command line option

nekipelov:
This option allows to use new git features
(example '--recurse-submodules' for issue #685). Or customize 'git
grep' per project with local .vimrc files.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -12,6 +12,10 @@ if !exists('g:fugitive_git_executable')
   let g:fugitive_git_executable = 'git'
 endif
 
+if !exists('g:fugitive_git_grep_options')
+  let g:fugitive_git_grep_options= '-n --no-color'
+endif
+
 " Section: Utility
 
 function! s:function(name) abort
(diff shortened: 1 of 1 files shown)


tpope:
You will need to make a very compelling argument for why `:Ggrep` needs a configuration option that `git grep` does not offer. Otherwise, your request should go to the maintainers of Git.

nekipelov:
New version of `git` offer the option `--recurse-submodules` but does not use by default. I want every `Ggrep` to work with `--recurse-submodules` for some projects. But not for all. In theory, in some projects I want to get  `Ggrep` with the option `--exclude-standard` or `--untracked`.

```
▶ git grep -h | fgrep submodules
    --recurse-submodules  recursively search in each submodule

▶ git --version                 
git version 2.16.2
```

tpope:
`:Ggrep` is a wrapper for `git grep`. If you want to call `git grep --recurse-submodules`, you can call `:Ggrep --recurse-submodules`. If I added this option, it would serve to confuse people why `:Ggrep` behaved differently than `git grep`. If you really don't want to type `--recurse-submodules`, send a patch to `git grep`, not `:Ggrep`.

jeffvandyke:
To add to this, if you wanted to `--recurse-submodules`, you could add a command remap, perhaps only for certain projects - see <https://stackoverflow.com/a/117221>.

On Git though, `git log --follow` comes to mind as a behavior that got a `.gitconfig` option in version 2.6 that can tell git to use `--follow` for all logs, and this sounds like something very similar. I'm thinking that a config option can be added for `grep.recurse-submodules` or something, and git configuration can be set by repository. I would highly recommend going down that route, because changing default behavior like that for some repos is something I find very useful.

aschrab:
Recent versions of git already support:

       submodule.recurse
           Specifies if commands recurse into submodules by default. This applies to all commands that have a
           --recurse-submodules option. Defaults to false.

PR #1002, opened 2018-03-22
Open files in a modifiable buffer, if possible

sodapopcan:
Sometimes I will either invoke, or jump to, the `Gstatus` window from an unmodifiable buffer which are almost always other plugins.  Then when I hit `o` to open a file, it opens the file in the plugin's window.
This is a mid-level inconvenience by me.

This patch will try and find a modifiable buffer to use if `winnr('#')` is `nomodifiable`.  If it can't, it will just use `winnr('#')` anyway.

This is definitely self-serving for my own plugin but happens from time to time with other plugins as well (like nerdtree).

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1424,11 +1424,19 @@ function! s:Edit(cmd,bang,...) abort
   let buffer = s:buffer()
   if a:cmd !~# 'read'
     if &previewwindow && getbufvar('','fugitive_type') ==# 'index'
+      let prev_winnr = winnr('#')
       if winnr('$') == 1
         let tabs = (&go =~# 'e' || !has('gui_running')) && &stal && (tabpagenr('$') >= &stal)
         execute 'rightbelow' (&lines - &previewheight - &cmdheight - tabs - 1 - !!&laststatus).'new'
-      elseif winnr('#')
+      elseif prev_winnr
         wincmd p
+        let skipbufs = ['nofile','help','quickfix']
+        while &previewwindow || index(skipbufs, &buftype) >= 0
+          wincmd w
+          if winnr() ==# prev_winnr
+            break
+          endif
+        endwhile
       else
         wincmd w
       endif


tpope:
Fugitive's historical buffers are not modifiable, and they are very much fair game. Also Netrw buffers. Probably others. I don't think this is a good option to key off of.

tpope:
Could maybe use `'buftype'`.

sodapopcan:
Considering your advice, it's a toss up between changing `while !&modifiable` to either `while &buftype !=# ''` or `while &buftype ==# 'nofile'`.  Both solve my problem, but I'm having a bit of trouble seeing the forest for the trees in terms of which is ultimately better.  If it were left up to me, I'd chose the former.

What I'm trying to accomplish is to provide a semi-decent in-vim rebase/cherry-pick/merge experience by providing a few "missing" fugitive mappings in my own buffer, namely `--continue`, `--skip`, and `--abort` (for their respective states).  For all other key-mapping needs, I want to delegate to `Gstatus`.  However, In the cases of `--continue` or `--skip`, these can involve jumps from my buffer to yours (heh).  This is where it gets messy.

I do have a plan b that isn't horrible and works, it's just not quite as nice of a user experience as I'm envisioning.

Thanks a lot for your feedback—ya got my brain all brainin'.

sodapopcan:
ugh, didn't mean to close 🙄

sodapopcan:
PS, before responding, I checked `&buftype` on every type of fugitive buffer I could think of and found they are all `''`.  My grokking (I hate and love that word) as to why you do this is beyond the scope of my prematurely submitting my previous comment.

tpope:
I'm not sure it's a great fit. My only use of `buftype` is for the `:Git!` and `:Gsplit!` buffers, and I can't quite remember why. Those would be good buffers to skip over though. Definitely want to skip `buftype=quickfix` and `buftype=help` too. Not sure about the other types. What does `:NERDTree` set it to?

sodapopcan:
Definitely help and quickfix!  Those are open far more often than nerdtree on my end but I wanted to give a general plugin example.

`:NERDtree` uses `buftype=nofile`, as does my plugin.  I think it's a commonly used one as it's suggested in Steve Losh's book.  Looks like `nofile` and `nowrite` only differ in how their names are handled.

Otherwise, I suggested skipping anything where `buftype` is set at all (or rather set to `''`) since I'd never want files opening in any type of buffer listed in `:h buftype`, although I'm admittedly still a little confused as to how `buftype=acwrite` works.  According to help, though, any buffer that has this option set is considered a `special-buffer`.

sodapopcan:
...which I've now realized doesn't quite work since it can open in `Gstatus` itself 😝.  It would have to be `&buftype !=# '' || &filetype !=# 'gitcommit'` which is starting to get ugly.  Or maybe `&buftype !=# '' || &readonly`?

I realize this is starting to get convoluted.  I'm willing to sit with some of these for a week or two and see how it goes.  Otherwise, I totally understand if you'd rather abandon this altogether.

tpope:
Just skip `&previewwindow`. You can't have another preview window open, but if you could, you'd definitely want to skip it too.

And yes, sitting on it a week or two sounds like a good idea. I'm already second guessing my recommendation to skip `:Git!` buffers. Maybe `buftype=nowrite` should be permissible? Things that correspond to "real files" feel like fair game. The other types seem pretty clearly skippable.

Hit me up when you have a solution you're happy with.

sodapopcan:
Soooo, I recently started a new job at a company where I'm always pairing and somehow found myself on a team that mostly uses RubyMine.  TL;DR, I haven't forgotten about this but I'm not getting to battle test much vim stuff recently in a collaborative environment.

In any event, I've checked out some popular/popularish (if not old) plugins (nerdtree, tagbar, gundo, unite, bufexplorer, minibufexplorer, buffergator,  taglist) which all `set buftype=nofile`.  Absolutely the idea is that anything that corresponds to a "real file" is fair game, though I feel something like `&previewwindow || index(['nofile', 'help', 'quickfix'], &buftype) >= 0` might be a little much for an edge case?

tpope:
It's more complicated than I hoped but that does sound lie the correct answer.

sodapopcan:
Yeah, more complicated than I hoped too.  I updated the PR anyway.  If you wanna merge then cool and if not then also cool :)

tpope:
Did this a bit more robustly, let me know how it works.

PR #1006, opened 2018-04-02
Optimize fugitive#foldtext

blueyed:
Only really relevant when using `set cursorline` [1], but good in
general after all.

1: https://github.com/vim/vim/issues/2773

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -3085,38 +3085,42 @@ augroup END
 function! fugitive#foldtext() abort
   if &foldmethod !=# 'syntax'
     return foldtext()
-  elseif getline(v:foldstart) =~# '^diff '
+  endif
+
+  let line_foldstart = getline(v:foldstart)
+  if line_foldstart =~# '^diff '
     let [add, remove] = [-1, -1]
     let filename = ''
     for lnum in range(v:foldstart, v:foldend)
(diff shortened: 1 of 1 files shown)


tpope:
That `modified|renamed` change is going to screw up non-English locales. Other than that looks good.

blueyed:
It only changes the quoting style compared to the original version, doesn't it?

tpope:
No, it replaces `[[:alpha:] ]+` with hard coded labels.

blueyed:
Are you mixing up lines?

This is from the diff:

```diff
-    cal map(lines,'s:sub(v:val, "^#\t%(modified: +|renamed: +)=", "")')
-    cal map(lines,'s:sub(v:val, "^([[:alpha:] ]+): +(.*)", "\\2 (\\1)")')
-    return getline(v:foldstart).' '.join(lines, ', ')
+    cal map(lines, "s:sub(v:val, '^#\t%(modified: +|renamed: +)=', '')")
+    cal map(lines, "s:sub(v:val, '^([[:alpha:] ]+): +(.*)', '\\2 (\\1)')")
+    return line_foldstart.' '.join(lines, ', ')
```

`modified|renamed` is on another line in the original code than the `[[:alpha:] ]+` already, and both are kept.
I could not figure out what the `([[:alpha:] ]+):` is supposed to match btw, but it appears to just swap `foo: bar` into `bar: foo`?

tpope:
You're right, I am mixing up lines.

Issue #1008, opened 2018-04-04
Strip out carriage returns


teoljungberg:
GitHub's new-ish [squash-and-merge button] seems to add carriage returns, that are visible in i.e `:Gedit`.

Example:

<img width="804" alt="screenshot 2018-04-04 14 52 42" src="https://user-images.githubusercontent.com/810650/38308571-d8a669d8-3817-11e8-99ed-9731d3da8d87.png">

Would it be possible to strip these out, or hide them somehow? Or is this a customization issue, where something in my local configuration mucks it up?

[squash-and-merge button]: https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits

tpope:
I've seen this from Windows users. Some GitHub workflow doing the same thing wouldn't surprise me. It would be nice to get rid of them!

tpope:
I take it back, the most recent incident was actually a Windows user ... getting squashed and merged by me.

teoljungberg:
I’m on macOS, so I agree that the issue lies within GitHub. So there’s probably nothing we can do there.

Would it be possible to strip them out? Or use a `fileformat` that silences them?

tpope:
`'fileformat'` is worthless for mixed files. Please do investigate stripping them out.

teoljungberg:
Gladly, any pointers on where to start looking inside fugitive?

teoljungberg:
@tpope The carriage returns have made their way back into fugitive.

Looking at https://github.com/tpope/vim-fugitive/blob/a9100fafb8c6deb187ee1cfaa37a4a02018d6c9a/autoload/fugitive.vim#L1503 it should still strip them out. Has there been any other refactorings like changing the filetypes that caused this?

teoljungberg:
If I remove `(exists(':keeppatterns') ? 'keeppatterns' : '')` from the command, it executes as expectedly removing the carriage returns.

What could be the culprit?

tpope:
Oh that is weird, I'm not sure. What if you flip it around so `keepjumps` comes first?

teoljungberg:
Sadly nothing, the issue still stands.

tpope:
What Vim version? Having trouble imagining any other explanation than a Vim bug.

teoljungberg:
I've tested this on both the version that ships with macOS and one built from source.

- 8.0.1283
- 8.1.473

tpope:
Can you point me at a commit for testing?

[...]

PR #1009, opened 2018-04-04
Strip out carriage returns

teoljungberg:
References #1008

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -2728,6 +2728,7 @@ function! s:BufReadObject() abort
           if lnum
             silent keepjumps delete_
           end
+          silent keepjumps 1,/^diff --git\|\%$/g/\r$/s///
           keepjumps 1
         endif
       elseif b:fugitive_type ==# 'blob'
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -2685,7 +2685,6 @@ function! s:BufReadObject() abort
     if b:fugitive_type !~# '^\%(tag\|commit\|tree\|blob\)$'
       return "echoerr ".string("fugitive: unrecognized git type '".b:fugitive_type."'")
     endif
-    let firstline = getline('.')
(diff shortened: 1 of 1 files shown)


plugin/fugitive.vim
```diff
@@ -2511,6 +2511,7 @@ function! s:ReplaceCmd(cmd,...) abort
   silent exe 'keepalt file '.tmp
   try
     silent edit!
+    silent exec '%g/'.nr2char(13).'/d'
```
teoljungberg: This is a very naive approach, I'd gladly take pointers!

tpope: First pointer is don't delete the entire line.

teoljungberg: 🤕 

Fixed in 04ee399 with a normal substitution.

teoljungberg: I also updated it to use the `c` s_flag to only remove the carriage returns if they exist.

teoljungberg:
Any thoughts on this @tpope? I've been running this branch locally since it was opened without any issues, if you can find any edge-cases for this implementation - or a better implementation - I'm all ears.

plugin/fugitive.vim
teoljungberg: Edit, not `c` but `e`

tpope:
Thinking something like `1,/^diff --git\|\%$/g/\r$/s///`, to avoid munging the diff.

teoljungberg:
TIL that you can search for a patterns for ex-commands. Thanks!

I rebased and squashed everything into two commits, one for the unused variable - the other for stripping out the carriage returns.

tpope:
Just noticed you're doing this to all sorts of buffers. Try it before https://github.com/tpope/vim-fugitive/blob/3bdac8b60383c0326991ba08bfd57d4e70925f3a/plugin/fugitive.vim#L2732. Based on the context you'll want to add a `keepjumps`.

PR #1015, opened 2018-04-10
Add hook at commit error

deresmos:
## Summary
A hook was added to use the error output when a commit error occurred.
Because I wanted to add pre-commit error part to qflist.
Error output such as pre-commit differs depending on the user, so we can make analytical program in each.
I think that users flexibly use various ways.

## Usage Example

1. `git clone https://github.com/potato4d/pre-commit-sample`
2. `cd pre-commit-sample && npm install`
3. Edit `index.js`. Change `const Name` to `var Name`
5. Run `:Gcommit`. Then only an error is displayed.
6. Add the following code to .vimrc.
```vim
function FugitiveCommitError(error, tmpfile)
  if a:error =~? 'pre-commit'
    call s:applyQflist(a:tmpfile)
  endif
endfunction

function! s:applyQflist(tmpfile)
  let errors = []
  let contents = readfile(a:tmpfile)

  let filepath = ''
  for content in contents
    if empty(content)
      continue
    endif

    if content =~? '^/' && filereadable(content)
      let filepath = content
      continue
    endif

    if content =~? '\v\s*[0-9]+:[0-9]+'
      let csp = split(content)
      call add(errors, filepath . ':' . csp[0] . ':' . join(csp[2:-2], ' ') . '. [' . csp[-1] . ']')
    endif
  endfor

  setlocal errorformat=%f:%l:%c:%m
  cgetexpr join(errors, "\n")
  copen
endfunction
```
7. After reloading `.vimrc`, once again `Gcommit`, qflist of the error part is displayed.

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -1136,6 +1136,9 @@ function! s:Commit(mods, args, ...) abort
       elseif error ==# '!'
         return s:Status()
       else
+        if exists('*FugitiveCommitError')
+          call FugitiveCommitError(error, errorfile)
+        endif
         call s:throw(empty(error)?join(errors, ' '):error)
       endif
     endif


tpope:
I'd rather have an autocommand than a function, but set that aside for now. Maybe we should just always use the quickfix list? Echoing one huge line has long been problematic. And we could match those trailing whitespace error messages from the example hook. The errorformat could be customizable. (You're doing some pre-processing, but I'm guessing it could be made into a pure errorformat pattern with `%P`.)

deresmos:
I also thought that it would be better to use autocommand than functions.
But is there a way to get the `errorfile` at the time of `Gcommit` from autocommand?
Since no method could be found, we wanted to get pre-commit error output, so we passed `errorfile` as an argument in the function.
When pre-commit error, I want to move to the file immediately, so use quickfix list.
If the cause of the error is not pre-commit, preprocessing can be prevented from being performed.

Thank you for the advice of errorformat. I will refer to it.

tpope:
This is possible now, see https://github.com/tpope/vim-fugitive/issues/1854#issuecomment-932374908. A filename containing the output of the command can be found at `result.file`.

PR #1025, opened 2018-05-03
#547 Add support for core.commentChar.

croesch:
Trying to solve #547 I would like you to have a look on my PR.

Please review the code style and my changes, especially the occurrences of `#` that I didn't touch!

I thought it would be necessary to escape the comment char (since it seems to be used in regular expressions) but I wasn't able to find a good way, anyway I created `s:repo_commentchar_escaped()` for that purpose. Can you help me with that? Surprisingly my special configuration `core.commentChar = $` seems to work after that PR, at least `-` is working in `:Gstatus`.

Consider this PR a draft and regarding to
>What takes you hours might take me minutes simply because I have both domain knowledge and a perverse knowledge of VimScript so vast that many would consider it a symptom of mental illness.

I'm looking forward reading your advices :)

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -93,7 +93,7 @@ endfunction
 function! s:recall() abort
   let rev = s:sub(s:buffer().rev(), '^/', '')
   if rev ==# ':'
-    return matchstr(getline('.'),'^#\t\%([[:alpha:] ]\+: *\)\=\zs.\{-\}\ze\%( ([^()[:digit:]]\+)\)\=$\|^\d\{6} \x\{40\} \d\t\zs.*')
+    return matchstr(getline('.'),'^'.s:repo_commentchar_escaped().'\t\%([[:alpha:] ]\+: *\)\=\zs.\{-\}\ze\%( ([^()[:digit:]]\+)\)\=$\|^\d\{6} \x\{40\} \d\t\zs.*')
   elseif s:buffer().type('tree')
     let file = matchstr(getline('.'), '\t\zs.*')
     if empty(file) && line('.') > 2
@@ -491,6 +491,14 @@ function! s:repo_user() dict abort
   return username.' <'.useremail.'>'
 endfun
 
+function! s:repo_commentchar() abort
+  return s:repo().config('core.commentChar')
+endfun
+
+function! s:repo_commentchar_escaped() abort
+  return '['.s:repo_commentchar().']'
+endfun
+
 function! s:repo_aliases() dict abort
   if !has_key(self,'_aliases')
     let self._aliases = {}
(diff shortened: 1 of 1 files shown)


tpope:
How do you plan to fix the syntax highlighting? This is a massive change for a partial solution. I think it'd be much simpler to just force `-c core.commentChar '#'` when we call `git status`.

Easiest way to escape is just by wrapping the character in square brackets.

tpope:
Where to begin. The concrete facts are that forcing `-c core.commentChar='#'` on `:Gstatus` would have no effect on any commit message. And `[[]` and `[]]` are valid.

It is my personal opinion that starting a commit message with `#issue-number` is a fool's errand. Does everyone who contributes to the project need to set that option too? Do they all need to go send pull requests to their respective editor tooling adding a bunch of special support? All so they can use some slightly different notation you just invented? Do you think this notation is so superior that the entire world will flock to it once it gains traction, and the short term pain is outweighed by the long term gain? Or do you just plan on slightly inconveniencing everyone around you indefinitely?

croesch:
I agree, a pragmatic solution.. I changed the obvious (to me) ones and left for example all `#\t` to use the comment char instead of dot.

What do you think about that set?

tpope:
The `Stage` functions are probably all fine to use `.`. Those buffers are 100% comments and there will never be a commit message to cause a false positive.

Repeatedly shelling out to `git config` is not an option. Do that in `foldtext` on Windows and you've rendered the editor completely unusable.

croesch:
OK .. I'll fix that.

What do you recommend when to call git config? Once per command / function or is that too often, too?

aschrab:
@croesch Are you aware of #689? I've been using that for a long time now without issues. It seems to handle this without needing to call `git config` at all.

tpope:
The Windows `system()` experience is unconditionally bad, and it can get painful even on UNIX if you're doing IO in the background. I think it needs to be fully cached and only called once per repo. This is a pain, and I highly recommend you try to avoid it. Consider `[[:punct:]]` in places where `.` seems too broad.

The patch looks like a blind search and replace. There's several places where something like `line ==# '#'` was changed to `line ==# '.'` which is obviously not going to work.

croesch:
@aschrab thanks, but unfortunately I was not ..

@tpope I'll add the caching.. It might look like but it is no *blind* search and replace. I still don't see ony occurence that looks like `==# '.'`. All `==# '#'` currently use the function to evaluate the comment char.. Can you give me one line number of those *several* places? I'll fix it then, too ..

tpope:
Searching for `==# '.` finds 5 matches.

tpope:
Rendered moot by the new `:Gstatus`.

PR #1027, opened 2018-05-06
Instaweb fallback

albfan:
Allow an url without handler to be show on instaweb

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -2423,23 +2423,6 @@ function! s:Browse(bang,line1,count,...) abort
   endtry
 endfunction
 
-function! s:github_url(opts, ...) abort
-  if a:0 || type(a:opts) != type({})
-    return ''
-  endif
-  let domain_pattern = 'github\.com'
-  let domains = exists('g:fugitive_github_domains') ? g:fugitive_github_domains : []
-  for domain in domains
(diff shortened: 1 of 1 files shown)


tpope:
If I read this correctly, it's backwards incompatible for the (admittedly small number of) users of the instaweb. Your principled aversion to seeing "github" in the source code is not nearly enough reason to break backwards compatibility.

tpope:
If I read this correctly, this breaks backwards compatibility for instaweb support. You're going to need to give a much better reason for breaking backwards compatibility.

albfan:
Just implement it to avoid suggestion without coding it. No real intention to modify

I will ask users of `:Gbrowse` if they feel natural to navigate to rhubarb plugin when they don't have github handler installed. Same as when you try to navigate to gitlab urls and fugitive just say "fugitive: 'origin' is not a supported remote"

No aversion at all, just see that warn in perspective as warn about sourceforge urls. Do you remember sourceforge? Time ago it was what today is github.

Thanks for the time to review this

tpope:
Believe me, I use the same Sourceforge analogy when complaining about those dumb plugin managers that do `:Plug 'username/reponame'`. The GitHub warning message is only to help with the extraction transition and will go away *eventually*, maybe in 3.0.

Opening https://github.com/tpope/vim-rhubarb was a lazy solution to the fact the warning can disappear in some setups without the user noticing. That last of the screen is super volatile.

Issue #1000, opened 2018-03-16
Feature Request: Gstatus on empty buffer


amadeus:
Not sure if this has been discussed before, however It would be rad if `:Gstatus` could work on empty buffer by inheriting the cwd, especially now that a `DirChanged` autocmd has been added to Vim since it could update from that event as well.

tpope:
Hasn't been discussed since `DirChanged`, but I'm unpersuaded. What happens if I `chdir` *out* of a repository? Am I supposed to manually delete all the commands one by one?

amadeus:
Sorry for the delayed response to this.

It may be tough to convince senpai, however I do think that `:Gstatus` should just be a command available anytime from anywhere, which would mean you actually would never have a need to clean it up. Within a CLI, if I you try to `git status` on a directory that isn't within git, you simply get the output `fatal: Not a git repository (or any of the parent directories): .git`.

I think it would be pretty useful if you could execute `:Gstatus` from anywhere - if it's from a buffer with a file, then it uses that file's git status - if there is none, it attempts to fallback to the working directory, if that still fails, it could either a) open that split with the `fatal` error message above, or just echo it to the cmd area. And if the buffer is not a file at all, it just attempts to use the current working directory with a similar output if it's a valid git directory or not.

tpope:
The difference is that Git provides one command everywhere, whereas Fugitive provides 30ish. I could see doing that for a general purpose SCM plugin, but *one* SCM just seems like a bridge too far.

Remind me why this is such a big deal? I've begrudgingly covered the `VimEnter` case. Is it just about being able to press Cmd-T in MacVim?

amadeus:
My usual workflow using MacVim, is that I generally use a window per project. Often times I will open up a MacVim window, cd into the directory of my project, and want to just do a quick `:Gstatus` to pull up where I am at. Currently I have to open up a file from the project before I can do this. 

The sort of inverse will also happen sometimes, I'll close out all open buffers from some work I am doing, but then want to call `:Gstatus` to monitor git things, but then need to re-open a file.

It's not a huge deal, but I do love the functionality of `:Gstatus` and often like to use it as a `hub` for what I am working on.

tpope:
[...]

PR #1033, opened 2018-05-28
Small tweak to improve fugitive-netrw interaction

Cimbali:
Sometimes netrw does not always set % correctly, e.g. for the first directory opened (such as `vim .` or `:e /some/path` with `set hidden`, whereas it is set after navigating within netrw), so this PR gets the directory opened in netrw from `b:netrw_curdir` instead. I reported the bug to vim, but this should work for all versions of netrw/vim.

From netrw's documentation (about *g:netrw_keepdir*):

	The current browsing directory is contained in
	b:netrw_curdir (also see |netrw-c|)

This means that instead of opening a random file in a repo to start fugitive we can open the directory instead (at least anyone using `set hidden`).

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -276,7 +276,7 @@ endfunction
 augroup fugitive
   autocmd!
   autocmd BufNewFile,BufReadPost * call fugitive#detect(expand('%:p'))
-  autocmd FileType           netrw call fugitive#detect(expand('%:p'))
+  autocmd FileType           netrw call fugitive#detect(fnamemodify(get(b:, 'netrw_curdir', @%), ':p'))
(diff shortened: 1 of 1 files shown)


tpope:
Never seen this bug, but as evidenced in my recent reply in #1000, this is a use case I care very much about. Let's use `fnamemodify(get(b:, 'netrw_curdir', @%), ':p')` because I worry some netrw change could break this.

Cimbali:
I checked with [the latest version of netrw](http://www.drchip.org/astronaut/vim/index.html#NETRW), where `b:netrw_curdir` always is the path of the current browsing directory (or latest directory opened in tree mode), but with display setup in tree-mode (`g:netrw_liststyle = 3`) the register `%` always contains `NetrwTreeListing`

tpope:
Ah, tree mode.

Applied by hand since there was a conflict.

PR #1039, opened 2018-06-07
Specify 'nofile' to status buffer and blame buffer

rhysd:
Since status buffer and blame buffer are not related to file and don't intend to be modified, buffer type should be 'nofile' I think. This is useful for me because I'm defining a mapping which closes 'nofile' buffers and it can also close fugitive's buffers when they are marked as 'nofile'.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -698,7 +698,7 @@ function! s:Status(bang, count, mods) abort
   try
     exe (a:mods ==# '<mods>' ? '' : a:mods) 'Gpedit :'
     wincmd P
-    setlocal foldmethod=syntax foldlevel=1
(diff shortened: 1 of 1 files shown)


tpope:
Setting `buftype=nofile` tells Vim the buffer name isn't a filename, which means it won't, for example, update the buffer name when changing directory. Since both kinds of buffer names really are real filenames, I don't think this is a good idea. You could get away with `buftype=nowrite`, I think.

rhysd:
Thank you for your quick reply.

Makes sense. I think I should update the logic with `nowrite` in my config.

And should we specify `nofile` in status buffer and blame buffer? If so, I'll update this PR.

tpope:
For `:Gstatus` definitely. For blame buffers it's debatable because they're temp files that you *could* write, but I say let's try it anyways.

PR #1047, opened 2018-06-29
Map 'p' to open in the preview window

odnoletkov:


tpope:
If we're going to do this we should do it for `:Gblame` too. There's also a conflict with `p` as `:Git add --path` in `:Gcommit` but we can sort that out later.

odnoletkov:
For ':Gblame' it semi-accidently doesn't jump to the chunk the same way as 'o' does. Actually I find it very handy since I more often just want to preview a commit header. If that's fine shall I make this jumping control explicit or leave as is?

tpope:
I think you're right about showing the top of the commit being more useful in `:Gblame`. But we should make the map explicitly do that rather than silently fail. Changing it to use `s:Edit` should do the trick.

Leave the now-obsolete maps in place in `:Gcommit` and I'll worry about further changes there later.

tpope:
I missed that you pushed the `:Gblame` change and did it independently, so that's taken care of. Next I tried your main `p` map, and it has a similar issue if you use it on something with a jump, which is a much bigger problem than for `:Gblame`. But long story short, I fixed it in 73c39195aee0c80f9ef3cd433ee315d72931864e. If you want to rebase your changes you're good to go.

odnoletkov:
👌

PR #1067, opened 2018-07-28
Correct the directory separator of Path in the windows environment

RyoSusami:
Patch for error from commit 07d30de in the windows environment.


tpope:
Thanks for looking out, can you just call `s:shellslash()` on the path and drop the conditional?

RyoSusami:
Thank you for pointing, I pushed the patch using s: shellslash().

tpope:
Thanks! This is clearly the correct logic, but can you clarify the error this was causing? So I can avoid future mistakes like this.

tpope:
Makes perfect sense, thanks!

PR #1079, opened 2018-08-18
s:ReplaceCmd: use bwipeout on the tmp bufname directly

blueyed:
When using "bash" as shell on Windows Vim, `tempname()` will return a
path with forward slashes, while fnamemodify+bufname use backward
slashes.

Using `s:tempname` for tmp results in mixed forward and backward
slashes (`C:/msys64/tmp\VIGCB01.tmp`).

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -1174,9 +1174,7 @@ function! s:ReplaceCmd(cmd) abort
     catch /^Vim\%((\a\+)\)\=:E302:/
     endtry
     call delete(tmp)
-    if fnamemodify(bufname('$'), ':p') ==# tmp
-      silent execute 'bwipeout '.bufnr('$')
-    endif
+    silent execute 'bwipeout '.s:fnameescape(tmp)
     silent exe 'doau BufReadPost '.s:fnameescape(fn)
   endtry
 endfunction


tpope:
That `if` statement was definitely there for a reason, I don't think we can just ditch it.

What is `:set shellslash?`. I thought it was all or nothing for forward slashes.

blueyed:
`&shellslash` is 0.
Vim appears to pick up `$SHELL` (which is `bash` via MSYS2).

I've found another problematic place in fugitive (`s:winshell`), which returns 1 in this case, because `&shellslash` is 0 - and then calls `cmd /c` in `s:TempCmd`.. :/

So the `if` is meant to make sure that the buffer has been created?  Or rather that it is still the last one?

tpope:
There's a special case in `tempname()` to use forward slashes if `'shellcmdflag'` starts with a `-`, so that things are less broken if you fail to set `'shellslash'`.

https://github.com/vim/vim/blob/111bbd61e96d1d3ee4bc07ae435fd97b88413aba/src/fileio.c#L7585-L7593

Try the `win32-shell` branch.

blueyed:
On master (with 8589246 already) and the win32-shell branch it does not work at all anymore, but silently.
`s:winshell` is 0 for me now, and this results in the fugitive URL not being matched in https://github.com/tpope/vim-fugitive/blob/1fa6e1c77b7fc75847582c5530aff071a9a0fea2/autoload/fugitive.vim#L426.
The fugitive URL looks like `fugitive:\\X:\.git\\` (before and after these changes).
I've tried using `s:winshell` also in `s:PlatformSlash`, but that did not help (but `s:Edit` appears to return the original filename then, i.e. there are two "edit" commands, but not for the fugitive URL).

But changing `s:Slash` appears to work:
```diff
diff --git i/autoload/fugitive.vim w/autoload/fugitive.vim
index dd8ddfc..621ddf2 100644
--- i/autoload/fugitive.vim
+++ w/autoload/fugitive.vim
@@ -82,7 +82,7 @@ function! s:warn(str) abort
 endfunction
 
 function! s:Slash(path) abort
-  if s:winshell()
+  if exists('+shellslash') && !&shellslash
     return tr(a:path, '\', '/')
   else
     return a:path
```

tpope:
Oh yeah that conditional is definitely wrong. I've changed it to the `'shellslash'` check but thinking about it I can't come with a reason not to *always* convert to forward slashes on Windows.

blueyed:
The original issue is not fixed on master (the buffer gets not wiped out still).
The bufname has backslashes, while tmp has forward slashes.
Should this be fixed on the win32-shell branch?  (currently conflicts with master)

blueyed:
win32-shell fixes it, regardless of what you pick for `s.winshell` there.

Issue #985, opened 2018-01-02
Display output during Gcommit


brendanator:
I have a pre-commit hook which runs units tests. These take quite a while and vim locks up while running them with no visible output

It would be great if the output could be displayed using `make` or dispatch

tpope:
Not possible with `:make` because of the way we need to intercept the editor invocation, but might be solvable with job support.

tpope:
Filing this under #1433. The branch there solves this.

atomdmac:
Sorry for the necromancy.  Happy to create a separate issue for this question if it would be helpful.

This doesn't seem to work for me.  I have `vim-dispatch` installed but doing after staging my files with `:G`, pressing `cc`, and saving my commit message, I'm unable to use `neovim` while the pre-commit script runs (with the output being displayed at the bottom).  My understanding is that having `vim-dispatch` installed should allow the output to display in a temporary buffer instead (a la `:G log`, etc.)

If I attempt to run `G! commit` instead, I get this message:

```
fugitive: aborting edit due to background operation.
error: There was a problem with the editor 'sh /var/folders/8k/74p75qrn5gx5yw8bh01j785r0000gn/T/nvimXaiYGU/5.sh'.
Please supply the message using either -m or -F option.
```

I assume I'm just doing something wrong but I'm not sure what that is.

tpope:
You're not doing anything wrong. There's no mechanism to commit in the background _and_ interactively edit the commit message in the foreground.

Issue #621, opened 2015-03-20
Print post-commit hook output to :Gcommit buffer 


huoxito:
So when I run `git commit` from shell I can see the output generated by the post-hook script fine, for example:

```
1 file inspected, no offenses detected
[master b96ab66] test
 Date: Fri Mar 20 11:43:17 2015 -0300
 1 file changed, 7 insertions(+), 1 deletion(-)
```

However when I commit inside vim via `:Gcommit` (which is what I do 90% of the times) it seems it always swallows any output on the post-commit hook:

```
"~/coding/example/.git/COMMIT_EDITMSG" 18L, 383C written

[master 60e3a57] test
 Date: Fri Mar 20 11:43:17 2015 -0300
 1 file changed, 7 insertions(+), 1 deletion(-)
```

I wonder if this is by design or if there's some sort of config / hack we could do to make sure the post-hook script results also show up on the Gcommit buffer? The post-hook script looks like this:

``` sh
#!/bin/sh
rubocop-git HEAD~1 HEAD
```

thanks

tpope:
My guess is that adding `2>&1` to the hook will fix it. The current implementation relies on capturing stderr and checking it for certain output, so it doesn't get displayed.

huoxito:
Appreciate the input @tpope I tried `rubocop-git HEAD~1 HEAD 2>&1` or even `echo huhu 2>&1` but it still behaves the same way posted above, I'm also not sure I understand why moving stderr to stdout would help (not that I know much about shell or anything at all about vim scripting), would the post-commit output be considered stderr? I'll try to keep digging need to know this better

tpope:
This should be fixed now. I didn't realize Git forced all hook output to stderr when I suggested that 4 years ago.

Issue #355, opened 2013-07-04
Interactive pre-commit hook waits indefinitely


kayleg:
When using a pre-commit hook which requires user input, the commit just hangs.

I am using https://github.com/nextbigsoundinc/Sprintly-GitHub hook which checks if the commit references a Sprint.ly item and prompts the user if it does not. 

The prompt never appears when committing using the Gstatus and Shift-C process

update: changed reference repository to correct one

lukeberry99:
I'm experiencing this issue at the moment, any work around for this?

tpope:
There's no mechanism for input, but it should fail now instead of hanging. I'm skeptical that interactive input can be shoehorned into the convoluted process `:Gcommit` uses to invoke the editor, but I'm open to ideas.

Issue #95, opened 2011-07-19
trimmed output of pre-commit hook


alfredodeza:
I took an initial look at issue #19 [0] but this seems a bit different.

I have to work with a set of pre-commit hooks that display errors in a few lines almost
always. So when something fails due to a pre-commit hook, I can't tell why because all
I am able to read is the first line of all the output (which does not include the _actual_ error).

It seems fugitive is using `messages`, and when I do check :messages I again see just one 
line.

Is there a way to see the whole output of a failure in a hook? 

I was using the latest version of the plugin (1.2) so I decided to checkout the latest changes from the
actual repo but I still found the same problem.

[0] https://github.com/tpope/vim-fugitive/issues/19

vnikk:
Hi. I got the same issue. When i do `:Gcommit` with some changes staged it just shows an empty bash screen (without any hook output). What would be the correct way to use those hooks with fugitive?

tpope:
I think this is *finally* sorted.

PR #1115, opened 2018-10-21
Show stderr output for commit hooks

Felixoid:
Currently, during pre/post commit hooks there is only silence in a console. With this patch issues #985, #621, #355 and #95 are going to be fixed.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -2011,7 +2011,11 @@ function! s:Commit(mods, args, ...) abort
       elseif a:args =~# '\%(^\| \)-\%(-interactive\|p\|-patch\)\>'
         noautocmd execute '!'.command.' 2> '.errorfile
       else
-        noautocmd silent execute '!'.command.' > '.outfile.' 2> '.errorfile
+        if executable('tee') && &shell =~# 'bash\|zsh'
+          noautocmd silent execute '!'.command.' > '.outfile.' 2> >(tee '.errorfile.')'
+        else
+          noautocmd silent execute '!'.command.' > '.outfile.' 2> '.errorfile
+        endif
       endif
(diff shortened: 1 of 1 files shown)


Felixoid:
You a right, this solution isn't so clear, but my research on how to capture and show stderr in vim script at the same time wasn't a success.

I've added a check for a shell as well.

Felixoid:
You a right, this solution isn't so clear, but my research on how to capture and show stderr in vim script at the same time wasn't a success.

I've added a check for a shell as well.

Felixoid:
Do you have, maybe, other ideas about how to figure out the lack of output here, @tpope ?

Felixoid:
Well, maybe we could merge it, @tpope ?

Felixoid:
Hey @tpope 
Could I hope that some action would be made here?

tpope:
Showing stderr but not stdout seems pretty arbitrary, no?

Felixoid:
During hooks execution every output is going to stderr. And the `outfile` will be shown later in 
```
if filereadable(outfile)
  for line in readfile(outfile)
    echo line
  endfor
endif
call fugitive#ReloadStatus()
return ''
```

tpope:
Okay, the thing I don't like about this is it forces everyone to get a flash of that `There was problem with the editor false.` message, in service of a tiny minority of users running a big slow command in their pre-commit hook. The fix here would be to move away from using `false` to something like `s:RebaseSequenceAborter()`, where we create a wrapper script that touches a tempfile that we can check for and then returns false.

As a more pragmatic fix, I've pushed a change to echo all error output except in the editing a message case. This won't solve the silent hang on a slow test suite problem, but it should solve most other use cases.

Issue #1058, opened 2018-07-20
Feature request: git hyper-blame


lanza:
Google's [depot_tools](https://commondatastorage.googleapis.com/chrome-infra-docs/flat/depot_tools/docs/html/git-hyper-blame.html) project has a `git hyper-blame` command that lets you ignore individual commits. EG a project wide formatting commit that destroys normal `git blame`'s usefulness without regularly hacking around it.

A project root file `.git-blame-ignore-revs` lets you permanently ignore all such troublesome commits. 

`Ghblame` would be a convenient mapping. 

I think I'll take a look into this at some point if nobody else is interested. Just figured I'd mention it incase somebody that knows the code base was interested.

tpope:
Doesn't look like it would require much more than tweaking a couple of regexps to account for the asterisks by the line numbers. I'm not super-keen on shipping out of the box support for a third party command, but I'm sure we can provide a hook or an option or something to make it simple to set up. First step would be to get a proof of concept working.

tpope:
It appears `--line-number` isn't supported. Without that, this is probably a non-starter.

mplatings:
In case it helps, Git 2.23 has git blame --ignore-revs-file built in. It's inspired by git-hyper-blame and also allows ignoring changes that split, join, add or remove lines. (I co-wrote it).
It might just work with vim-fugitive automatically if you set blame.ignoreRevsFile in your .gitconfig file.

tpope:
Yep, already supported by Fugitive. See also discussion in #1316.

PR #1119, opened 2018-11-13
Support custom blame command, e.g. hyper-blame

afrojer:
This patch allows a user to set a global variable which specifies the
name of the git command used when fugitive wants to git-blame. To
variable name is: "g:fugitive_blame_cmd" and you can use it in your
vimrc file like so:

if executable('git-hyper-blame')
  let g:fugitive_blame_cmd = 'hyper-blame'
endif

NOTE: the custom blame command's output must match git's base
blame command output.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -2990,7 +2990,8 @@ function! s:Blame(bang, line1, line2, count, mods, args) abort
       call s:throw('unsupported option')
     endif
     call map(a:args,'s:sub(v:val,"^\\ze[^-]","-")')
-    let cmd = ['--no-pager', 'blame', '--show-number']
+    let bcmd = get(g:, 'fugitive_blame_cmd', 'blame')
+    let cmd = ['--no-pager', bcmd, '--show-number']
     if a:count
       let cmd += ['-L', a:line1 . ',' . a:line1]
     endif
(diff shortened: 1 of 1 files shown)


tpope:
Does this actually let you jump to commits? In #1058 I determined `--line-number` wasn't supported. Did that change, maybe?

afrojer:
The depot_tools git-hyper-blame has not changed, and `--show-number` is still unsupported. I have a commit that adds this (and other basic git-blame) support to git-hyper-blame, but it might take a bit to get it merged.

The idea would be to require whatever external git-blame command you run to support basic git-blame arguments and produce equivalent output. This seems like the least amount of work for the vim-fugitive plugin. Although to get colors to look right, you may need to change the regex to ignore the extra `*` that git-hyper-blame puts in.

tpope:
Well I guess that's acceptable. I'll even allow the regexp tweak for the `*`.

Let's go with `_command` for symmetry with other options.

blueyed:
Interesting, just did https://github.com/tpope/vim-fugitive/pull/1120 to add support for `--first-parent`.

Given this PR here it might make sense to allow for a list, so that you could also use this for default options maybe?  (so that you could use `['blame', '--first-parent']` with this new setting)

tpope:
As much as I love arrays of arguments, I think allowing `'blame --first-parent'` would be more intuitive. The implementation could call `split(value, ' ')`.

Another note: the use of `blame` inside `:Gbrowse` is pretty low level, and I don't think overriding it is necessary or wise.

tpope:
I have unified `:Git blame` and `:Gblame`. So my new position is that it would be weird if `:Git blame` called `hyper-blame`, but having `:Git hyper-blame` reuse `:Git blame`'s internals would be totally cool.

Issue #1125, opened 2018-11-29
GPush/GPull is stuck when it's asking for a new server fingerprint for ssh


iamkarlson:
I'm pretty sure it's gonna be end by timeout or whatever but it's really confusing when you are pushing/pulling on the new machine.

tpope:
Neovim?

iamkarlson:
yep

iamkarlson:
do you need any logs or other stuff?

tpope:
Neovim breaks `:!`. You can work around it by installing dispatch.vim and vim-dispatch-neovim.

PR #1126, opened 2018-11-29
Added information about broken bang in the readme

iamkarlson:


justinmk:
Adding notes in various plugins about Neovim bugs is just noise. And when we fix the bug then there will be a bunch of outdated notes floating around.

iamkarlson:
> when we fix

Sometimes this is not gonna happen.

justinmk:
The bug will be fixed eventually.

iamkarlson:
@justinmk I'm not so sure about that. It can be just replaced with :term or :shell or any other stuff. So far I saw that neovim promised to have the same API with some additional features but then --wait and --remote are gone, bang is broken. It's not a personal insult to any of neovim developers but just a fact. There's no any update since Apr 18 for the issue with a broken bang and I believe that lots of people experienced it. What is their the last step before neovim hangs? fugitive. Let's make users' life easier instead of increasing sunk costs of getting acquainted with internal mechanics of fugitive.

iamkarlson:
#1125

tpope:
As frustrating as this issue is, it does seem to be Google-able enough that bug reports for the password prompt have stopped coming in. Hopefully #1125 does the same for SSH key prompts. A FAQ entry might still be justifiable, but it should be something specific like "Why does `:Gpush` hang?", not vague and ominous like "Some functionality is broken".

README.markdown
```diff
@@ -96,6 +96,11 @@ to open after any grep invocation:
 
     autocmd QuickFixCmdPost *grep* cwindow
 
+>Gpush or Gpull hangs in the Neovim. What can I do?
+
+Neovim breaks :! and fugitive cannot process request from the command line. You can work around it by installing dispatch.vim and vim-dispatch-neovim.
+
```
justinmk: "Neovim breaks :!" is an uncharitable wording. And not precise enough. And as I already mentioned, gVIm has the same behavior in some circumstances. It only applies to interactive commands. So at least mention "interactive programs" or something like that. It's not helpful to start rumors like ":! is broken on Neovim".

tpope:
I added a FAQ entry addressing the password problem.

Issue #323, opened 2013-03-26
Diffing staged changes gives "E21: Cannot make changes, 'modifiable' is off: fileencoding=UTF-8"


admk:
# Issue

Diffing staged changes gives "E21: Cannot make changes, 'modifiable' is off: fileencoding=UTF-8", when modeline is on and using `# vim: set fileencoding=UTF-8 :` for python scripts.
# How to reproduce
1. `set modeline`.
2. create an empty repository.
3. commit a python script with the following lines at the top of the file:

``` python

    #!/usr/bin/env python
    # vim: set fileencoding=UTF-8 :
```
1. make more changes to the file.
2. stage it with `:Gst` and `-`.
3. `dd` on the staged changes.
# Screenshot

![Screen Shot 2013-03-26 at 17 54 12](https://f.cloud.github.com/assets/238821/304498/37d0c7f4-963e-11e2-921e-772ea403b181.png)
# Vim version info

```
VIM - Vi IMproved 7.3 (2010 Aug 15, compiled Mar 26 2013 12:50:46)
MacOS X (unix) version
Included patches: 1-875
Compiled by ko@ko-mba.local
Huge version without GUI.  Features included (+) or not (-):
+arabic          +file_in_path    +mouse_sgr       +tag_binary
+autocmd         +find_in_path    -mouse_sysmouse  +tag_old_static
-balloon_eval    +float           +mouse_urxvt     -tag_any_white
-browse          +folding         +mouse_xterm     -tcl
++builtin_terms  -footer          +multi_byte      +terminfo
+byte_offset     +fork()          +multi_lang      +termresponse
+cindent         -gettext         -mzscheme        +textobjects
-clientserver    -hangul_input    +netbeans_intg   +title
+clipboard       +iconv           +path_extra      -toolbar
+cmdline_compl   +insert_expand   -perl            +user_commands
+cmdline_hist    +jumplist        +persistent_undo +vertsplit
+cmdline_info    +keymap          +postscript      +virtualedit
+comments        +langmap         +printer         +visual
+conceal         +libcall         +profile         +visualextra
+cryptv          +linebreak       +python          +viminfo
+cscope          +lispindent      -python3         +vreplace
+cursorbind      +listcmds        +quickfix        +wildignore
+cursorshape     +localmap        +reltime         +wildmenu
+dialog_con      -lua             +rightleft       +windows
+diff            +menu            +ruby            +writebackup
+digraphs        +mksession       +scrollbind      -X11
[...]

PR #1136, opened 2018-12-22
Set nomodeline before loading blob

blueyed:
I have found this because of an error through an invalid modeline
(`E488: Trailing characters: et'`), which prevented fugitive from
loading the blob (via `:Gdiff`).

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -1575,8 +1575,8 @@ function! fugitive#BufReadCmd(...) abort
       elseif b:fugitive_type ==# 'stage'
         call s:ReplaceCmd([dir, 'ls-files', '--stage'])
       elseif b:fugitive_type ==# 'blob'
-        call s:ReplaceCmd([dir, 'cat-file', b:fugitive_type, rev])
         setlocal nomodeline
+        call s:ReplaceCmd([dir, 'cat-file', b:fugitive_type, rev])
       endif
     finally
       keepjumps call setpos('.',pos)


tpope:
This throws out the baby with the bathwater. If you're using modelines, you'd certainly want them for index files too. Why does it prevent it from loading the blob? What step is failing?

See also #323.

blueyed:
Ok, you're right.
What is the purpose of the `setlocal nomodeline` afterwards?

It fails here: https://github.com/tpope/vim-fugitive/blob/70170d825c320e37b1aee785a718171885092d71/autoload/fugitive.vim#L1365 - when commenting or try/catching this it fails here: https://github.com/tpope/vim-fugitive/blob/70170d825c320e37b1aee785a718171885092d71/autoload/fugitive.vim#L1354 (that might also just be an issue with the reported location).
But it makes sense that the `edit` fails, which appears to handle modelines also.

Should it get retried with `nomodeline` then?  (reporting the original error)

tpope:
The `setlocal nomodeline` was added in response to #323, and it's a pretty crappy workaround quite frankly.

I would expect a failing `doau BufReadPost` to cause the subsequent steps not to run, but it happens after the buffer has already been loaded, so I wouldn't it expect to see an empty buffer. Is that what you're seeing? An empty buffer?

blueyed:
No, the window is closed after the error gets displayed.
But have not tested with a minimal setup.
However, without the `doau` it clearly throws at the `edit`.
I assume the window is closed due to https://github.com/tpope/vim-fugitive/blob/70170d825c320e37b1aee785a718171885092d71/autoload/fugitive.vim#L1363.

tpope:
Okay, so a first good step (which I just took) is to disable modelines for the `edit` command, and let the error happen on `BufReadPost`.

blueyed:
Ok, but the overall result is still the same for me, i.e. the window is not opened.
When commenting the `silent exe 'doau BufReadPost '.s:fnameescape(fn)` it then complains about not being a fugitive buffer (via my statusline I assume).

tpope:
Long story short, moving the `doautocmd` outside of the function and directly into the original autocommand seems to fix it.

Issue #797, opened 2016-05-22
RFE: support for the interactive rebase


mcepl:
I need to reorganize a huge branch with plenty of commits on many files (porting of `M2Crypto` to Python 3) and in the end in order to make it at least some order in it, I exploded each large commits touching many files with this script (called `git-split-by-file`):

``` sh
#!/bin/sh                                                                       
set -xeu                                                                        

if [ $# -gt 0 ] ; then                                                          
    MSG=$(cat "$1")                                                             
else                                                                            
    if [ -f ".git/rebase-merge/message" ] ; then                                
                MSG=$(cat .git/rebase-merge/message)                            
        else                                                                    
                read -e -p "Enter the merge commit message: " MSG               
                if [ -z "$MSG" ] ; then                                         
                        exit 1                                                  
                fi                                                              
    fi                                                                          
fi                                                                              

function add_one_file () {                                                      
    FILE="$1"                                                                   
    git add "$FILE"                                                             
    echo -n "$FILE: ${MSG}"|git commit -F -                                     
}                                                                               

git reset HEAD~                                                                 
for f in $(git diff |lsdiff --strip 1) ; do                                     
    add_one_file "$f"                                                           
done
```

So, splitting is not a problem, but now I have a branch with not 40, but 120 commits. I have cut it down back to 50 by combining commits touching one file into one, but now I would like to cut the branch even further down: by combining many tiny but related commits into one. The best workflow I found (and I don't like it) is to have all those commits on gitlab opened in the Firefox, in the terminal to have opened `git rebase -i` window and picking and combining commits to be squashed together.

Which is quite crazy. What I am dreaming about is to have the `git rebase --interactive` todo list opened in a vim buffer, and then be able to open from inside of this buffer underlying commits in `Gdiff` window, so I can see exactly which files I am picking for squash. This would IMHO hugely improve efficiency of my `git rebase -i` session.

What do you think?

tpope:
I would love this. But how to implement? It see no way to invoke `git rebase` interactive functionality without it invoking an editor, and invoking Vim from inside Vim is not a great use experience.

odnoletkov:
@mcepl I think most of what you described is already possible. With cursor somewhere on the commit hash in the rebase to-do buffer in Vim:
`K` is essentially `git show`
`:Gedit <CTRL-R CTRL-W>` opens fugitive buffer for commit (also see `:Gsplit`/`:Gvsplit`/`:Gtabedit`/`:Gpedit`)

[...]

PR #1139, opened 2018-12-26
Support Grebase --continue/--edit-todo editor invocation

rbong:
As discussed in #797 

`copen` was decided as the method of feedback for the user on `Grebase --continue` because `cwindow` obfuscates user feedback.

The exact method for `--edit-todo` discussed was not used. You don't need to copy the file, edit it, and move it back. You just need to edit the file. The only difference between this and invoking `--edit-todo` via the command line is that the commit hashes are not contracted and expanded. This requires using `git rebase--interactive --expand-ids` and `git rebase--interactive --shorten-ids`, which are internal helpers and can't be relied on to remain publicly accessible.

`Grebase --edit-todo` does not yet support window mods but it easily can. Let me know if you think this is needed now.

I also believe that it would be useful to have `--continue` and `--edit-todo` rebase autocompletion. This is another thing that can easily be added if you think it should be part of this PR.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -2336,7 +2336,7 @@ let s:common_efm = ''
 
 function! s:Merge(cmd, bang, mods, args) abort
   let mods = substitute(a:mods, '\C<mods>', '', '') . ' '
-  if a:cmd =~# '^rebase' && ' '.a:args =~# ' -i\| --interactive\| --edit-todo'
+  if a:cmd =~# '^rebase' && ' '.a:args =~# ' -i\| --interactive'
     return 'echoerr "git rebase --interactive not supported"'
   endif
   let [mp, efm] = [&l:mp, &l:efm]
@@ -2398,6 +2398,8 @@ function! s:Merge(cmd, bang, mods, args) abort
     elseif a:cmd =~# '^rebase' && ' '.a:args =~# ' --continue' && s:HasRebaseCommitCmd()
       cclose
       return mods . 'Gcommit --amend'
+    elseif a:cmd =~# '^rebase' && ' '.a:args =~# ' --edit-todo'
+      return mods . 'Gsplit ' . s:fnameescape(b:git_dir . '/rebase-merge/git-rebase-todo') . ' | setlocal bufhidden=wipe nobuflisted'
     endif
   endif
   let qflist = getqflist()
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -2308,6 +2308,13 @@ function! s:RemoteComplete(A, L, P) abort
   return join(matches, "\n")
 endfunction
 
+function! s:HasRebaseCommitCmd() abort
+  if !filereadable(b:git_dir . '/rebase-merge/amend') || !filereadable(b:git_dir . '/rebase-merge/done')
+    return 0
+  endif
+  return get(readfile(b:git_dir . '/rebase-merge/done'), -1, '') =~# '^[^e]'
+endfunction
+
 function! fugitive#Cwindow() abort
   if &buftype == 'quickfix'
     cwindow
@@ -2345,6 +2352,7 @@ function! s:Merge(cmd, bang, mods, args) abort
           \ . '%+ECannot %.%#: Your index contains uncommitted changes.,'
           \ . '%+EThere is no tracking information for the current branch.,'
           \ . '%+EYou are not currently on a branch. Please specify which,'
+          \ . '%+I%.%#git rebase --continue,'
           \ . 'CONFLICT (%m): %f deleted in %.%#,'
           \ . 'CONFLICT (%m): Merge conflict in %f,'
           \ . 'CONFLICT (%m): Rename \"%f\"->%.%#,'
@@ -2383,10 +2391,13 @@ function! s:Merge(cmd, bang, mods, args) abort
     execute cdback
   endtry
   call fugitive#ReloadStatus()
-  if empty(filter(getqflist(),'v:val.valid'))
+  if empty(filter(getqflist(),'v:val.valid && v:val.type !=# "I"'))
     if !had_merge_msg && filereadable(b:git_dir . '/MERGE_MSG')
       cclose
       return mods . 'Gcommit --no-status -n -t '.s:shellesc(b:git_dir . '/MERGE_MSG')
+    elseif a:cmd =~# '^rebase' && ' '.a:args =~# ' --continue' && s:HasRebaseCommitCmd()
+      cclose
+      return mods . 'Gcommit --amend'
     endif
   endif
   let qflist = getqflist()


autoload/fugitive.vim
```diff
@@ -2328,8 +2348,15 @@ let s:common_efm = ''
...
 
 function! s:Merge(cmd, bang, args) abort
-  if a:cmd =~# '^rebase' && ' '.a:args =~# ' -i\| --interactive\| --edit-todo'
+  if a:cmd =~# '^rebase' && ' '.a:args =~# ' -i\| --interactive'
     return 'echoerr "git rebase --interactive not supported"'
+  elseif a:cmd =~# '^rebase' && ' '.a:args =~# ' --edit-todo'
+    if !filereadable(b:git_dir . '/rebase-merge/interactive')
+      return 'echoerr "no rebase in progress"'
+    elseif !filereadable(b:git_dir . '/rebase-merge/git-rebase-todo')
+      return 'echoerr "could not open todo file"'
```
tpope: Instead of inventing our own errors, I think we should fall back to the usual execution path and let Git complain.

autoload/fugitive.vim
```diff
@@ -2328,8 +2348,15 @@ let s:common_efm = ''
...
+    endif
+    return 'Gsplit ' . s:fnameescape(b:git_dir . '/rebase-merge/git-rebase-todo')
```
tpope: I just pushed mods support. Sorry about the merge conflict.

autoload/fugitive.vim
```diff
@@ -2386,6 +2413,13 @@ function! s:Merge(cmd, bang, args) abort
     if !had_merge_msg && filereadable(b:git_dir . '/MERGE_MSG')
       cclose
       return 'Gcommit --no-status -n -t '.s:shellesc(b:git_dir . '/MERGE_MSG')
+    elseif a:cmd =~# '^rebase' && ' '.a:args =~# ' --continue'
+      if s:has_rebase_commit_cmd()
+        cclose
+        return 'Gcommit --amend'
+      elseif !empty(getqflist())
+        return 'call fugitive#Copen()'
```
tpope: Rather than using `:copen`, I think what you want to do is add something like `%+I  git rebase --continue` to the error format, so that the window pops open whenever instructions are displayed.

autoload/fugitive.vim
```diff
@@ -2308,17 +2308,37 @@ function! s:RemoteComplete(A, L, P) abort
   return join(matches, "\n")
 endfunction
 
-function! fugitive#Cwindow() abort
+augroup fugitive_rebase
+  autocmd!
+  autocmd FileType gitrebase setlocal bufhidden=wipe
```
tpope: No good, you'll wipe out the buffer in a command line `git rebase -i` invocation if the user navigates away. Chain this onto the returned `Gsplit` line.

autoload/fugitive.vim
```diff
@@ -2308,17 +2308,37 @@ function! s:RemoteComplete(A, L, P) abort
...
+augroup END
+
+function! s:has_rebase_commit_cmd() abort
```
tpope: I'm trying to use camel case for all functions going forwards.

autoload/fugitive.vim
rbong: That's a good idea. I will execute `--edit-todo` with git editor `false` and that should do it.

autoload/fugitive.vim
rbong: Yup, that does it.

rbong: Sorry, this got lost when editing the history.

rbong: Alright, this fix is back in now.

rbong:
I've done some more testing and the code that adds informational rebase messages to the quickfix causes the commit window not to open. I'm pushing up a fix momentarily that updates the code which checks that there are no errors in the quickfix list by excluding informational lines.

tpope:
This will result in a "problem with editor false" style error in the quickfix list on `--edit-todo`, correct? I think the check for it should be moved up to avoid that. Other than that looks good.

tpope:
I'll deal with that later.

rbong:
It will result in that error in the quickfix list, but the quickfix list will not be opened. However, we can avoid that error by simply changing the editor "true" rather than "false" when "--edit-todo" is passed in. I'm pushing that up now.

I am not entirely sure if this is the approach you were thinking of, but it preserves error checking by running the `git rebase --edit-todo` command while also keeping the "problem with editor false" error out of the quickfix window in case the user opens it.

Also, I want to make sure you know that the same behaviour will happen on "--continue". The editor error will be in the quickfix window, but the quickfix window will not automatically open if there is a commit message to edit. Causing an error using `$GIT_EDITOR` is the only way I know of to pause the rebase so we can edit commit messages. I assumed that was okay because `%-Gerror:%.%#false` was in the list of error formats, but if it's not okay and there's some way to get around actually having this error in the quickfix list let me know.

tpope:
Yeah the error format should filter out the false message now that I think about it, but I don't see any reason to clobber the quickfix list at all in the successful `--edit-todo` case. If by "preserves error checking" you mean it prevents it from happening outside a rebase, I think the `filereadable()` check is fine for that. So it skips `:make` if we can edit the sequence file but runs it if we can't for the error message.

tpope:
Any thoughts on how we might make `:Grebase --continue` invoke `:Gcommit` in the case of a resolved merge conflict?

rbong:
I'll have to look into it more

amadeus:
`:Grebase --continue` from a resolved merge conflict would be very nice :X kinda weird to have to jump out of the `:Grebase` flow when dealing with conflicts (a very common thing for me, heh)

tpope:
It's driving me nuts. If someone can figure out the conditions of *when* we need to call `git commit`, then I can probably figure out the how. The answer should be in the `git rebase --continue` source code.

rbong:
Rebase [detects unresolved conflicts by looking for unstaged changes](https://github.com/git/git/blob/master/builtin/rebase.c#L1254). Even if you are not in the middle of a merge conflict resolution, it will complain that you are if you have unstaged files.

Similarly, if you have any staged files and you run `git rebase --continue`, git enters into the commit editor whether you previously had a merge conflict or not. I cannot find the specific code for this but I've confirmed it through experimentation. If you need better than that, let me know. The flow is a little complex and the easiest way to find out what's going on is probably by compiling git with debug information and using gdb to trace it.

Minor note: if there are conflicts on `git rebase --continue` it will leave a `MERGE_MSG`... however, deleting this and running `--continue` results in the same behaviour (including the correct commit message), so it is probably not good to rely on this file at all.

rbong:
My last post was correct but I have actually confirmed it in the source code now. There is only one minor special case to consider.

The main points about how it works are below, but more information is available here if you need to dig into it:

- just to show the flow up to the relevant point: when running `git rebase --continue`, `git rebase--interactive --continue` is called internally and [drops into sequencer_continue](https://github.com/git/git/blob/e902e9bcae2010bc42648c80ab6adc6c5a16a4a5/builtin/rebase--interactive.c#L241-L243), which then performs validation of [the internal cache, the options](https://github.com/git/git/blob/e902e9bcae2010bc42648c80ab6adc6c5a16a4a5/sequencer.c#L3969-L3973), and the [todo file](https://github.com/git/git/blob/e902e9bcae2010bc42648c80ab6adc6c5a16a4a5/sequencer.c#L3975-L3976). All of this validation can be put off by running `rebase --continue` after editing the commit as far as I can tell. The only relevant validation that occurs before getting to `rebase--interactive` is checking if there are unstaged changes, shown in my post above
- The sequencer [immediately tries to commit any unstaged changes](https://github.com/git/git/blob/e902e9bcae2010bc42648c80ab6adc6c5a16a4a5/sequencer.c#L3977-L3978).
- The sequencer again [checks for any unstaged changes](https://github.com/git/git/blob/e902e9bcae2010bc42648c80ab6adc6c5a16a4a5/sequencer.c#L3832-L3833) and fails if they exist, although it should not get to this point
- If `rebase-merge/amend` exists, it goes down a [long branching path](https://github.com/git/git/blob/e902e9bcae2010bc42648c80ab6adc6c5a16a4a5/sequencer.c#L3837-L3932). I haven't dug too deep into this because we're already handling the existence of this file.
- If [there are no uncommitted changes](https://github.com/git/git/blob/e902e9bcae2010bc42648c80ab6adc6c5a16a4a5/sequencer.c#L3835) and no errors handling the `rebase-merge/amend` file, [the sequencer goes back to continuing the rebase](https://github.com/git/git/blob/e902e9bcae2010bc42648c80ab6adc6c5a16a4a5/sequencer.c#L3934-L3941)
- At this point, [the changes are committed](https://github.com/git/git/blob/e902e9bcae2010bc42648c80ab6adc6c5a16a4a5/sequencer.c#L3943-L3945) and if there are any errors the sequencer fails. `run_git_commit` only checks if there is currently an interactive rebase, not for this special case, so it is the same as the other rebase cases we're currently handling
- Finally, [any remaining interactive rebase state files are cleaned up and the normal continue workflow is returned to](https://github.com/git/git/blob/e902e9bcae2010bc42648c80ab6adc6c5a16a4a5/sequencer.c#L3946-L3961). Again, we're already handling these files and I haven't dug too deep into what happens when they're cleaned up

Main takeaways:
- As suspected, rebase checks to see if there was a merge conflict that needs resolving by looking at staged files. If there are unstaged files, it considers it an unresolved conflict, and if there are staged files, it commits them to resolve any conflicts. There doesn't appear to be any particularly special considerations
- There may be special considerations if files such as `rebase-merge/amend` exist already and there are staged files, but the way we're currently handling those files may be sufficient. Practical testing is needed before I go digging into that workflow

I hope there's something in there to help implement this. It's certainly beyond my capabilities to actually add this behaviour to the plugin without some serious investment.

tpope:
I can't imagine I covered every use case but it's now working for the ones I bothered to test.

PR #1140, opened 2018-12-27
Use `Make` in Merge when it is availiable

bennyyip:


tpope:
This breaks the `Gcommit` functionality for editing the commit message later in the function which assumes synchronous execution.

I don't really think this should be asynchronous, as wrecking the work tree while editing can lead to undesired behavior.

PR #1142, opened 2018-12-28
Move `Unstaged` below `Staged` header

teoljungberg:
`git status` displays "Changes to be committed" above "Changes not
staged for commit" and "Untracked files". This keeps fugitive.vim
consistent with that layout.


tpope:
This is what Magit does and it makes a ton of sense. Staging files is by far the more common operation than unstaging so they should be closer to the cursor.

PR #1168, opened 2019-01-10
Use git-grep column feature introduced in git 2.19.0

llorens:


tpope:
I like your version check API enough to put my own spin on it. Thanks!

llorens:
Very nicely done! Thanks!

PR #1179, opened 2019-01-23
Fix s:Cd() usage in Gblame

johnmiked15:
Most usages in commit da3b2f32 (Support tab local working directories)were updated to use this pattern, but Gblame missed the boat. Without this, Gblame will change the root git tree but won't change back to the previous location.


tpope:
When you say "won't change back to the previous location", are you hypothesizing or is this behavior you have witnessed? Based on my reading of the code, you've introduced a bug where the returning cd runs only in the blame window, not the original window.

johnmiked15:
I can reproduce the behavior. If my working directory is below the git root, and I run :Gblame, the window I was working in changes its current working directory to the git root, and does not change back to where it was, even after closing the :Gblame window.

I verified this change does change the working directory of the main window back to what it started as after running the git blame in the root.

johnmiked15:
Your changes from the commit I linked all follow the same pattern. `s:Cd()` will actually change the directory and return a statement that is later used to go back to the previous working directory, typically in the finally statement for the try block that does the initial `cd`. All usages of `s:Cd()` were updated to use this pattern, except the one in `:Gblame`.

johnmiked15:
@tpope, have you had a chance to look at this any closer or reproduce the original issue locally?

I’ve reproduced it on several machines, and confirmed this PR fixes the problem.

autoload/fugitive.vim
```diff
@@ -3550,24 +3550,14 @@ function! s:Blame(bang, line1, line2, count, mods, args) abort
...
       let temp = error.'.fugitiveblame'
       if &shell =~# 'csh'
         silent! execute '%write !('.basecmd.' > '.temp.') >& '.error
       else
         silent! execute '%write !'.basecmd.' > '.temp.' 2> '.error
       endif
-      if exists('l:cwd')
-        execute cd s:fnameescape(cwd)
-        unlet cwd
-      endif
```
tpope: This was here because running in the `finally` causes an `lcd` to happen after the vertical split, which is too late. I recommend moving the `finally` block up here, and adding a second `try`/`endtry` block.

johnmiked15: Sure, I'll take care of it.

PR #1186, opened 2019-01-29
Add `g:fugitive_replace_status_on_commit` option

ddrscott:
Option defaults to disabled/unset.
When enabled, the pre-Dec 2019 behaviour of :Gcommit taking over
:Gstatus will occur.

Related to https://github.com/tpope/vim-fugitive/commit/a32c301f516cf64e5a8e28ad106923e8cee5cda3

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -2495,6 +2495,10 @@ function! s:Commit(mods, args, ...) abort
           execute mods 'keepalt edit' s:fnameescape(msgfile)
         elseif a:args =~# '\%(^\| \)-\w*v' || mods =~# '\<tab\>'
           execute mods 'keepalt -tabedit' s:fnameescape(msgfile)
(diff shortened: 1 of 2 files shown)


tpope:
To answer your question, the original behavior was to avoid the confusing situation of having 2 near identical looking windows with vastly different purposes. The fact it spared the need to close the window - a single `q` keystroke - was a minor gain, but the fact it made it impossible to keep the window if you weren't done with it was a major drawback. Not to mention having `:Gcommit` (and not just the maps) behave differently in different contexts was its own brand of confusion.

The redesign means I can ditch the practice and I am doing so with great prejudice. It's dead. While I'm sensitive to even minor backwards compatibility breakages, the downsides here are so inconsequential (at least to my eye) that I'm not really interested in a `g:fugitive_pry_it_from_my_cold_dead_hands` option.

One of the many changes to `:Gstatus` is that it now accepts a count as height - `:10Gstatus` for a 10 line window - and you can give `:0Gstatus` to skip the split entirely. I think `:0Gcommit` would be a reasonable addition (and you could map it as well).

PR #1187, opened 2019-01-31
Move filetype to after maps to allow user maps

idbrii:
The easiest way for users to define their own maps in Gstatus buffers is
using a ftplugin/fugitive.vim file with a `b:fugitive_type != 'index'`
guard. Setting the filetype after defining the maps allows users to
maniuplate "dangerous" maps and otherwise customize mappings.

For example, in ~/.vim/after/ftplugin/fugitive.vim:

    if !exists('g:loaded_fugitive') || b:fugitive_type != 'index'
        finish
    endif

    " Use s to stage *and* unstage changes: shadows -.
    nmap <buffer> <silent> s -
    xmap <buffer> <silent> s -

    " I hit this by accident.
    nunmap <buffer> X

Alternatively, the buffer settings and maps could be moved to a ftplugin/fugitive_status.vim file (is that how it worked for the old status?).

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -1521,7 +1521,6 @@ function! fugitive#BufReadStatus() abort
 
     set nomodified readonly noswapfile
     silent doautocmd BufReadPost
-    set filetype=fugitive
     setlocal nomodifiable
     if &bufhidden ==# ''
       setlocal bufhidden=delete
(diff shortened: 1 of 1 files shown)


tpope:
We probably should make an ftplugin but this is a no brainer first step.

It should be safe to omit the `b:fugitive_type` check. I'd like to get rid of that variable entirely someday.

PR #1189, opened 2019-02-01
Allow tab as a modifier for Gstatus

### LINKED ISSUE #880: Support :tab Gstatus
It would be nice if `:tab Gstatus` would open the status window in a new tab.

This could be implemented by using `<mods>` in the command mapping, if the Vim version is sufficiently recent (>= 7.4.1898).

theryangeary:
Add tab as an option for <mods> to Gstatus. Closes #880.


odnoletkov:
`:tab G` works es expected for me without this

theryangeary:
Yeah I don't know what I did to convince myself this was necessary or why I wasn't able to do it before but it sure does work without it!

Issue #834, opened 2016-09-08
Autocommands do not fire for files named `index`


brennen:
I have an autocommand to do some stuff.  I noticed that it fails on files named `index`, I think because fugitive.vim has its own autocommand for `index{,.lock}` that triggers `BufReadIndex()`, which in turn does something to prevent further processing of the event. I'm kind of hazy on how.

Background: I have some blog entries named like `foo/archives/2016/9/8`, and others named like `foo/archives/2016/9/7/index`.  I couldn't figure out why [an autocommand](https://github.com/brennen/bpb-kit/blob/master/home/.vimrc#L389) to set `filetype` was failing on the latter, and figured it must be because I had written the pattern wrong.  Eventually I realized that the function wasn't getting called at all for the `index` files, even with a pattern of `*`.

I dug around in the output of `:autocmd`, which led me to:

```
fugitive_files  BufReadCmd
    index{,.lock}
              if fugitive#is_git_dir(expand('<amatch>:p:h')) |   exe s:BufReadIndex() | elseif filereadable(expand('<amatch>')) |   read <amatch> |   1delete | endif
    fugitive://**//[0-3]/**
              exe s:BufReadIndexFile()
    fugitive://**//[0-9a-f][0-9a-f]*
              exe s:BufReadObject()
```

It turns out that if I start vim without fugitive, the autocmd fires as expected.

Here's a sample `.vimrc`:

```
set nocompatible

filetype off

set rtp+=~/.vim/bundle/Vundle.vim/
call vundle#rc()

" let Vundle manage Vundle - required:
Plugin 'VundleVim/Vundle.vim'
Plugin 'tpope/vim-fugitive'

au BufReadPost,BufNewFile * echom "my autocommand fired"
```

This can be tested with `:e index` and `:e some_other_file`.

This is pretty edge-case-y, and not a serious impediment to my work or anything, but it does feel like a thing that shouldn't really happen.

Thanks!

tpope:
I think `doautocmd` is the way to go. It's imperfect but at 2 lines of code I wouldn't really call it complicated.

brennen:
I wasn't sure if it was appropriate to fire both "missing" commands every time - seems like strictly speaking `BufNewFile` at least oughta be conditional on whether the file's new, and I'm not sure whether there are other commands that're blocked here.

Anyhow I can throw together a PR at some point.

tpope:
[...]

PR #1196, opened 2019-02-10
BufRead(Pre|Post), BufNewFile autocmds for files named "index" (#834)

brennen:
- Adds BufReadPre and BufReadPost around read operation if the file is not in a `.git/` and already extant.
  - Adds BufNewFile for new files.

I'm not sure if this covers every corner case, but it solves my original problem and appears to behave as expected with `.git/index` files.

Should fix: https://github.com/tpope/vim-fugitive/issues/834

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -271,14 +271,21 @@ augroup fugitive
         \ endif |
         \ let b:undo_ftplugin = get(b:, 'undo_ftplugin', 'exe') . '|setl inex= inc='
 
+  " Handle index files in Git directories, emulating normal sequence of
+  " autocommands if a file named `index` is found elsewhere:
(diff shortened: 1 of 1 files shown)


brennen:
::sigh::

Clearly didn't read commit message line length requirement closely enough.  Got to run out the door right now, but I'll squash commits and resubmit.

tpope:
If no autocmds exist, `doautocmd` outputs a warning message. Let's address this by adding `silent`.

I need to update that commit hook to give a more useful error. My personal objection would be I don't want `tpope#959` in my history. My style is `Closes https://...` in the body. (Or you can omit it entirely, this doesn't really seem worthy of a paper trail.)

brennen:
Moved to #1197.

Issue #959, opened 2017-10-26
vim-fugitive breaks CTRL-^ alternate file editing for files named index


gsauthof:
How to reproduce:

1. go to an empty directory; create 2 text files named index and foo
2. vim index
3. :e foo
4. Ctrl-6 (switches back to index)
5. Ctrl-6 (expected behaviour: switches back to foo, actual behaviour: nothing happens)

I tested with current vim-fugitive on Fedora 26 with a minimal config (i.e. just pathogen and vim-fugitive bundles enabled). When disabling vim-fugitive the alternate file editing works as expected.

Possibly related issue: Autocommands do not fire for files named `index` #834

brennen:
@gsauthof I have a PR in for #834 - see #1196.  It doesn't appear to fix this issue, although I wonder if it might still be a related problem.  I'm not quite sure how `C-^` is implemented under the hood.  Perhaps because the `index` file read happens inside an autocmd using `read <amatch>`, some state is not getting set correctly.

I wonder if some variable could, perhaps, just be set here...

tpope:
I bet `keepalt read` will fix it, if you wanna go for a two-in-one.

brennen:
Sure enough.  Added to PR.

PR #1197, opened 2019-02-11
Fix autocmds & alternate file handling for "index"

brennen:
This commit tweaks the existing autocommand for files matching `index{,.lock}` to:

  - Fire BufReadPre and BufReadPost on either side of the read operation for files not in a `.git/` and already extant.
  - Fire BufNewFile for new files not in a `.git`.
  - Add `keepalt` to the `read`, thus allowing alternative file operations such as Ctrl-^ to work correctly.

Has been tested against:

  - BufReadPost,BufNewFile autocommands
  - A basic test case for alternate files where one is named `index`

Closes https://github.com/tpope/vim-fugitive/issues/834

Closes https://github.com/tpope/vim-fugitive/issues/959

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -271,14 +271,21 @@ augroup fugitive
         \ endif |
         \ let b:undo_ftplugin = get(b:, 'undo_ftplugin', 'exe') . '|setl inex= inc='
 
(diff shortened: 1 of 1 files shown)


tpope:
Lol I'd rather have a 52 character summary than a damn ampersand.

PR #1216, opened 2019-02-23
Reblame at HEAD from dirty lines

odnoletkov:
HEAD blame seems to be much more useful than index blame. You can continue following evolution of the line with further jumps while before you got struck on 'not yet committed' line if it was already added to the index.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -3868,7 +3868,7 @@ function! s:BlameJump(suffix) abort
   let commit = matchstr(getline('.'),'^\^\=\zs\x\+')
   let suffix = a:suffix
   if commit =~# '^0\+$'
-    let commit = ':0'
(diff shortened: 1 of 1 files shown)


tpope:
I mean if I was going to change anything I would make this a no-op. The line number is going to be wrong in all but the most reductive cases. If you wanna see a commit pick a committed line!

odnoletkov:
I want to see a 'previous commit' for the line like these maps do for non-dirty lines. Practical use case is to find a proper fixup/squash target for the line. Blame at HEAD clearly shows what commit you want to amend for modified lines and sometimes helps with added lines as well. Do you know a better way?

I missed the problem with line numbers. It can be fixed by applying offset calculated from the diff output similar to how it's done in other places.

tpope:
I still think this is weird but I guess it doesn't hurt anything.

PR #1243, opened 2019-04-04
Allow --abbrev as :Gblame argument

jsit:
I'd like to narrow the width of the :Gblame content by showing only 8 characters of the SHA-1 string instead of the default 10. Adding `--abbrev` to the permitted flags lets me do this.


tpope:
The default isn't 10, it's auto-selected based on the size of your repository. I'm fine to add it to the pile of undocumented options.

PR #1244, opened 2019-04-04
Add documentation for --relative-date in :Gblame

jsit:
`--relative-date` is allowed in `:Gblame` but not documented; this fixes that.

### DIFF
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -186,9 +186,10 @@ that are part of Git repositories).
                                                 *fugitive-:Gblame*
 :Gblame [flags]         Run git-blame on the file and open the results in a
                         scroll bound vertical split.  You can give any of
-                        ltfnsewMC as flags and they will be passed along to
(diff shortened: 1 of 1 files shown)


tpope:
On second thought, there are several undocumented options, and you picked the one that doesn't really work right to document. The syntax highlighting and column width are wrong. I'm reverting.

jsit:
Hm, curious to know what vim-fugitive is doing to the git-blame output then, since the raw output of `git-blame --relative-date` aligns the columns correctly.

tpope:
Fugitive hides the trailing whitespace before the line number and closing parentheses, which is necessary for the fixed width date format but causes a jagged edge with the variable width relative date format.

A solution to both this and the syntax problem would be to match a fixed width relative date field. Something like `\%(\d\+ \w\+ ago\)\@=[[:alnum:] ]\{14\}`. That is, assert that something like `30 minutes ago` is up ahead, then match exactly 14 characters.

PR #1248, opened 2019-04-13
Abbreviate commit SHA1s for `:Grebase -i`

teoljungberg:
Rather than generating the full 40 character SHA1, this PR changes
the rebase todo file to use the abbreviated kind.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -2851,12 +2851,7 @@ function! s:RebaseEdit(cmd, dir) abort
         endif
         let shortened_sha = sha[0:s:sha_length]
         let b:rebase_shas[shortened_sha] = sha
-        let new_rebase_todo[i] = substitute(
-              \ new_rebase_todo[i],
-              \ sha,
-              \ shortened_sha,
-              \ '',
-              \ )
+        let new_rebase_todo[i] = substitute(new_rebase_todo[i], sha, shortened_sha, '')
       endif
     endfor
     call writefile(new_rebase_todo, rebase_todo)
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -2844,7 +2844,7 @@ function! s:RebaseEdit(cmd, dir) abort
     let shas = {}
 
     for i in range(len(new))
-      if new[i] =~ '^\l\+\s\+[0-9a-f]\{3,\}\>'
+      if new[i] =~# '^\l\+\s\+[0-9a-f]\{3,\}\>'
         let sha = matchstr(new[i], '\v[a-f0-9]{5,40}')
         if !sha_length
           let sha_length = len(s:TreeChomp(a:dir, 'rev-parse', '--short', sha))
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -2993,7 +2993,7 @@ function! s:RebaseClean(file) abort
 
     let sha = matchstr(new[i], '\v[a-f0-9]{5,40}')
     let rebase_shas = getbufvar(a:file, 'fugitive_rebase_shas')
-    if len(sha) && type(rebase_shas) == v:t_dict && get(rebase_shas, sha)
+    if len(sha) && type(rebase_shas) == type({}) && get(rebase_shas, sha)
       let new[i] = substitute(new[i], sha, rebase_shas[sha], '')
     endif
   endfor
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -2840,14 +2840,18 @@ function! s:RebaseEdit(cmd, dir) abort
 
   if filereadable(rebase_todo)
     let new_rebase_todo = readfile(rebase_todo)
+    let s:sha_length = 0
 
     for i in range(len(new_rebase_todo))
       if new_rebase_todo[i] =~ '^[' . join(values(s:rebase_abbrevs), '|') . ']'
         let sha = matchstr(new_rebase_todo[i], '\v[a-f0-9]{5,40}')
+        if !s:sha_length
+          let s:sha_length = len(call('system', ['git rev-parse --short ' . sha]))
+        endif
         let new_rebase_todo[i] = substitute(
               \ new_rebase_todo[i],
               \ sha,
-              \ sha[0:7],
+              \ sha[0:s:sha_length],
               \ '',
               \ )
       endif
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -2840,23 +2840,23 @@ function! s:RebaseEdit(cmd, dir) abort
 
   if filereadable(rebase_todo)
     let new = readfile(rebase_todo)
-    let s:sha_length = 0
-    let b:rebase_shas = {}
+    let sha_length = 0
+    let shas = {}
 
     for i in range(len(new))
       if new[i] =~ '^\l\+\s\+[0-9a-f]\{3,\}\>'
         let sha = matchstr(new[i], '\v[a-f0-9]{5,40}')
-        if !s:sha_length
-          let s:sha_length = len(call('system', ['git rev-parse --short ' . sha]))
+        if !sha_length
+          let sha_length = len(call('system', ['git rev-parse --short ' . sha]))
         endif
-        let shortened_sha = sha[0:s:sha_length]
-        let b:rebase_shas[shortened_sha] = sha
+        let shortened_sha = sha[0:sha_length]
+        let shas[shortened_sha] = sha
         let new[i] = substitute(new[i], sha, shortened_sha, '')
       endif
     endfor
     call writefile(new, rebase_todo)
   endif
-  return a:cmd . ' +setlocal\ bufhidden=wipe ' . rebase_todo
+  return a:cmd . ' +setlocal\ bufhidden=wipe\|' . escape('let b:fugitive_rebase_shas = ' . string(shas), ' ') . ' ' . rebase_todo
 endfunction
 
 function! s:Merge(cmd, bang, mods, args, ...) abort
@@ -2992,13 +2992,13 @@ function! s:RebaseClean(file) abort
     let new[i] = substitute(new[i], '^\l\>', '\=get(s:rebase_abbrevs,submatch(0),submatch(0))', '')
 
     let sha = matchstr(new[i], '\v[a-f0-9]{5,40}')
-    if len(sha) && get(b:rebase_shas, sha)
-      let new[i] = substitute(new[i], sha, b:rebase_shas[sha], '')
+    if len(sha) && get(b:fugitive_rebase_shas, sha)
+      let new[i] = substitute(new[i], sha, b:fugitive_rebase_shas[sha], '')
     endif
   endfor
   if new !=# old
     call writefile(new, a:file)
-    unlet b:rebase_shas
+    unlet b:fugitive_rebase_shas
   endif
   return ''
 endfunction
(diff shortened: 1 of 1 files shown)


autoload/fugitive.vim
```diff
@@ -2825,7 +2825,25 @@ let s:rebase_abbrevs = {
...
+  if filereadable(rebase_todo)
+    let new_rebase_todo = readfile(rebase_todo)
+
+    for i in range(len(new_rebase_todo))
+      if new_rebase_todo[i] =~ '^[' . join(values(s:rebase_abbrevs), '|') . ']'
+        let sha = matchstr(new_rebase_todo[i], '\v[a-f0-9]{5,40}')
+        let new_rebase_todo[i] = substitute(
+              \ new_rebase_todo[i],
+              \ sha,
+              \ sha[0:7],
```
teoljungberg: This feels naive, but the performance loss of using `git rev-parse --short SHA` for each commit was too great.

Any ideas how to abbreviate the SHA in a better way?

autoload/fugitive.vim
```diff
@@ -2825,7 +2825,25 @@ let s:rebase_abbrevs = {
...
 
 function! s:RebaseEdit(cmd, dir) abort
-  return a:cmd . ' +setlocal\ bufhidden=wipe ' . s:fnameescape(a:dir . '/rebase-merge/git-rebase-todo')
+  let rebase_todo = s:fnameescape(a:dir . '/rebase-merge/git-rebase-todo')
+
+  if filereadable(rebase_todo)
+    let new_rebase_todo = readfile(rebase_todo)
+
+    for i in range(len(new_rebase_todo))
+      if new_rebase_todo[i] =~ '^[' . join(values(s:rebase_abbrevs), '|') . ']'
```
teoljungberg: I guess this could go the other way check that the line isn't a comment? What would be preferred?

tpope: I would do something like `^\l\+\s\+[0-9a-f]\{3,\}\>'. The single letter check erroneously includes `label` and omits commands that may be added in the future.

autoload/fugitive.vim
tpope: `git rebase--interactive --shorten-ids` would probably do the trick. Otherwise, do the first line by shelling out and then use the length of that SHA to truncate the rest.

tpope:
I don't want to abbreviate SHAs unless we can also restore them afterwards. `git rebase--interactive --expand-ids` might do the trick or you could save them in a dictionary.

teoljungberg:
@tpope Where did you find `--shorten-ids` and `--expand-ids`? I can't find them in the rebase man page.

https://git-scm.com/docs/git-rebase

autoload/fugitive.vim
teoljungberg: I updated to get the length of the sha once in 1600c108179fa02250e49b78e6f15b74381d6c2d

autoload/fugitive.vim
teoljungberg: Done in bcc0e4d

teoljungberg:
I updated the implementation to expand the SHA1s afterwards by keeping them in a dictionary, as suggested.

tpope:
`rebase--interactive` is a helper utility. Now that I look at it, the options I suggested are too new to be relied upon.

Use a buffer local variable for the dictionary. It's quite possible to have 2 rebases in 2 repositories happening at the same time.

teoljungberg:
@tpope Gotcha, I'll look into `rebase--interactive` for the kicks of it.

I updated to use a buffer local variable in 1927e6c, and then did some refactoring to get consistency with the rest of the code base in 7ed27b7, 478c618, and e197828. Are there are other irregularities you'd like me to tackle?

teoljungberg:
The buffer local variable didn't seem to do the trick, the buffer for `RebaseClean` doesn't seem to be the same one as for `RebaseEdit`.

```
Error detected while processing function <SNR>64_RebaseClean:
line   10:
E121: Undefined variable: b:rebase_shas
E116: Invalid arguments for function get(b:rebase_shas, sha)
```

teoljungberg:
We could work around the issue with editing the todo file first, then setting the variables for the SHA dictionary and SHA length, and after computing both of them editing the previous file.

Here's the diff for it:

```diff
diff --git a/autoload/fugitive.vim b/autoload/fugitive.vim
index a02d2e2..8b4dad1 100644
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -2829,20 +2829,22 @@ function! s:RebaseEdit(cmd, dir) abort
 
   if filereadable(rebase_todo)
     let new = readfile(rebase_todo)
-    let s:sha_length = 0
+    execute 'edit ' . rebase_todo
+    let b:sha_length = 0
     let b:rebase_shas = {}
 
     for i in range(len(new))
       if new[i] =~ '^\l\+\s\+[0-9a-f]\{3,\}\>'
         let sha = matchstr(new[i], '\v[a-f0-9]{5,40}')
-        if !s:sha_length
-          let s:sha_length = len(call('system', ['git rev-parse --short ' . sha]))
+        if !b:sha_length
+          let b:sha_length = len(call('system', ['git rev-parse --short ' . sha]))
         endif
-        let shortened_sha = sha[0:s:sha_length]
+        let shortened_sha = sha[0:b:sha_length]
         let b:rebase_shas[shortened_sha] = sha
         let new[i] = substitute(new[i], sha, shortened_sha, '')
       endif
     endfor
+    execute 'edit #'
     call writefile(new, rebase_todo)
   endif
   return a:cmd . ' +setlocal\ bufhidden=wipe ' . rebase_todo
```

But, that feels brittle and hacky. Any other suggestions?

tpope:
Something like this maybe:

    return a:cmd . ' +setlocal\ bufhidden=wipe\|' . escape('let b:fugitive_rebase_shas = ' . string(shas), ' ') . ' ' . rebase_todo

teoljungberg:
Worked out great, pushed!

teoljungberg:
Is there anything outstanding @tpope ?

tpope:
`s:RebaseClean()` takes a filename, operating completely independently of the current buffer. I did it this way because I find `BufDelete` and `VimLeavePre` to be a little unpredictable. Using a variable from the current buffer violates that. See if you can make it work with `getbufvar()`.

teoljungberg:
Good point, done in e74359a to use `getbufvar()`.

autoload/fugitive.vim
```diff
@@ -2971,9 +2991,16 @@ function! s:RebaseClean(file) abort
...
+
+    let sha = matchstr(new[i], '\v[a-f0-9]{5,40}')
+    let rebase_shas = getbufvar(a:file, 'fugitive_rebase_shas', {})
+    if len(sha) && get(rebase_shas, sha)
+      let new[i] = substitute(new[i], sha, rebase_shas[sha], '')
+    endif
   endfor
   if new !=# old
     call writefile(new, a:file)
+    unlet b:fugitive_rebase_shas
```
tpope: Won't work.

autoload/fugitive.vim
```diff
@@ -2971,9 +2991,16 @@ function! s:RebaseClean(file) abort
   let new = copy(old)
   for i in range(len(new))
     let new[i] = substitute(new[i], '^\l\>', '\=get(s:rebase_abbrevs,submatch(0),submatch(0))', '')
+
+    let sha = matchstr(new[i], '\v[a-f0-9]{5,40}')
+    let rebase_shas = getbufvar(a:file, 'fugitive_rebase_shas', {})
+    if len(sha) && get(rebase_shas, sha)
```
tpope: Drop the third `getbufvar()` argument (it's a relatively recent addition) and instead add a `type()` check to the conditional.

autoload/fugitive.vim
teoljungberg: Good catch, done in 5b75c48

autoload/fugitive.vim
teoljungberg: Done in d0d0ca5, I went with comparing `type(rebase_shas) == v:t_dict`. I hope that's compatible enough, or do you think we should go with `type(rebase_shas)  == type({})`?

teoljungberg: `v:t_dict` was introduced in vim8, so I changed it to be compatible in 5e64be4

autoload/fugitive.vim
```diff
@@ -2971,6 +2991,12 @@ function! s:RebaseClean(file) abort
   let new = copy(old)
   for i in range(len(new))
     let new[i] = substitute(new[i], '^\l\>', '\=get(s:rebase_abbrevs,submatch(0),submatch(0))', '')
+
+    let sha = matchstr(new[i], '\v[a-f0-9]{5,40}')
+    let rebase_shas = getbufvar(a:file, 'fugitive_rebase_shas')
+    if len(sha) && type(rebase_shas) == type({}) && get(rebase_shas, sha)
```
tpope: You need to use `has_key()`, not `get()`, or this will only work for SHAs that convert to a nonzero integer. `'deadbeef'` is false in a boolean context.

autoload/fugitive.vim
```diff
@@ -2837,7 +2837,27 @@ let s:rebase_abbrevs = {
...
+    let sha_length = 0
+    let shas = {}
+
+    for i in range(len(new))
+      if new[i] =~ '^\l\+\s\+[0-9a-f]\{3,\}\>'
+        let sha = matchstr(new[i], '\v[a-f0-9]{5,40}')
+        if !sha_length
+          let sha_length = len(call('system', ['git rev-parse --short ' . sha]))
+        endif
+        let shortened_sha = sha[0:sha_length]
```
tpope: `string[0:2]` returns characters 0, 1, and 2, for a length of 3. Use `strpart()` to avoid the off by one error.

Also, you're including a newline in the count, and you're making assumptions about the current directory. `s:TreeChomp(a:dir, 'rev-parse', '--short', sha)` should address both of these.

autoload/fugitive.vim
teoljungberg: Done in 3b863ab

autoload/fugitive.vim
teoljungberg: Went with `strpart` in 38bc5a0

And used `TreeChomp` in 3243b24

teoljungberg:
Have you had a chance to look at this again @tpope ?

teoljungberg:
Is there anything holding this PR from getting re-reviewed? I think I applied all of the changes that were requested.

tpope:
Use `=~#`, and add `\C` to the other regexps for symmetry.

Given the number of bugs I spotted just looking at the code, I'm worried about the ones I missed, to be frank. What steps have you taken to test this?

teoljungberg:
I've used this for rebasing since I opened the PR without any hiccups, that includes squash commits, fixup commuts, etc that could change the layout of the rebase todo file.

Can you point out the bugs you found, and let's see if we can address them?

tpope:
I've pointed out the ones I found, it's the ones I haven't that worry me.

Fix the case sensitivity thing (more of a linter concern than a practical bug) and I suppose I'm ready to merge.

teoljungberg:
I added using `\C`  to the regexpes in 2728271d0c48ff993bed2f6191fd5eb040c0a281, aswell as rebased the branch.

tpope:
I just force pushed this away because I took "rebased" to mean "cleaned up the history" and not "moved it". I don't see a way to redo to please resubmit.

teoljungberg:
My bad! I reopened the PR as https://github.com/tpope/vim-fugitive/pull/1260

tpope:
Eh this one is at least half my fault. 🙃

Issue #776, opened 2016-04-04
:Ge and friends act strangely when work dir is separate from git dir


h3xx:
## Symptoms:

When operating on files in a git repo where the working directory is separate from the git directory, some commands act strangely.
- `:Gw` fails with `fugitive: cannot determine file path`
- `:Ge` opens the parent directory of the git directory in netrw
-  `:Gvdiff` fails with `"[parent dir of git dir]" Illegal file name`
- `:Glog` successfully loads the history into the quickfix buffer, but closes the file buffer, instead showing a commit as if I had typed <kbd>cc</kbd>. Pressing <kbd>Enter</kbd> in the quickfix window updates the commit buffer, but does not show the file at that revision.
- `:Gstatus` shows correct working directory status, but pressing <kbd>D</kbd> over a modified file shows no differences. Pressing <kbd>Enter</kbd> opens the file in the git dir's parent directory (NOT the work dir).
## Steps to Reproduce:
1. Use the following shell commands to setup an environment:
   
   mkdir TEMP
   cd TEMP
   git clone git://github.com/tpope/vim-fugitive.git
   mkdir workdir
   cd workdir
   git init --separate-git-dir ../vim-fugitive/.git
   git checkout .
   echo 'changed line' >>README.markdown
2. Open README.markdown in vim.
3. Attempt to run any of the commands listed.
## Version information:
- vim-fugitive 57afba5bdd49b59d57485f90aa0cf3f3a6c47f63 (2016-03-23)
- vim 7.4 (patches 1-1530)
- git 2.8.0 (all git commands work as expected in working directory)

Verified under a clean-room setting with only vim-fugitive installed.

tpope:
Can confirm all if this.

systemmonkey42:
I just chased down a similar problem, this time with 'Ggrep' reporting `fugitive: no work tree`.

calling `git config --add core.worktree $(pwd)`

seemed to calm it down.   Even though there is a file called '.git' with a 'gitdir:' link, it can't (or wont) locate the work tree.

This, of course, doesn't help if I want to maintain multiple work tree's.   Git grep will find .git, identify the incorrect work tree, and probably not do the Right Thing ™

h3xx:
Strangely, fugitive seems to be able to find everything and operate normally when in a submodule of a project with a separate git dir.

I.e. the superproject is like the one I described, and it has a submodule that's cloned using `submodule update --init`. When working on files in that submodule, fugutive works just fine.

[...]

PR #1262, opened 2019-06-04
Fix issues when git-dir is a symlink

### LINKED ISSUE #1253: It breaks again when .git is a symlink.
For example, `GStatus` shows the files in `.git` folder rather than the source files, `GVdiff` also does not work.

brookhong:
Test steps:

1. open a file in a repo whose .git is a symbol link, make some modification on the file.
1. run `:Gvdiff` in vim, which should show correct diff.
1. run `:Gstatus` in vim, which should list current file under `Unstaged`.
1. in the `FUGITIVE` buffer for `Gstatus`, go to line of current file, press `dv`, which should show correct diff.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -308,7 +308,17 @@ function! fugitive#Prepare(...) abort
   call s:PreparePathArgs(cmd, dir, !exists('explicit_pathspec_option'))
   let args = join(map(copy(cmd), 's:shellesc(v:val)'))
   if empty(tree) || index(cmd, '--') == len(cmd) - 1
-    let args = s:shellesc('--git-dir=' . dir) . ' ' . args
+    let git_dir = s:shellesc('--git-dir=' . dir)
+    if exists('g:git_dir_links')
+      for d in keys(g:git_dir_links)
+        if d == dir
+          let worktree_for_git_link = substitute(g:git_dir_links[d], '\.git$', '', '')
+          let git_dir = git_dir . ' ' . s:shellesc('--work-tree=' . worktree_for_git_link)
+          break
+        endif
+      endfor
+    endif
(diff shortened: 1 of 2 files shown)


tpope:
One thing I didn't realize when I responded to #1253 (which I discovered looking into #776) is that `git config core.worktree $(pwd)` - the standard way to tell Git where the worktree lives for a repo - is enough to get Fugitive to work with a symlinked Git dir. If that's good enough, you might want to walk away before burning any more time on this.

As far as making it work *without* `core.worktree` goes, this is a good first attempt, but it doesn't fully solve the problem, only mitigates it. If you have 2 different `.git` symlinks pointing at the same directory, for example, Fugitive will get very confused.

The "right" way to go about this is to flip it: put the symlink in `b:git_dir`, and perhaps use a dictionary to cache where it links to. [The `.git/` handling portion of `FugitiveFind()`](https://github.com/tpope/vim-fugitive/blob/41a734ef49c78cfdb97426cda811ee99c39e011a/autoload/fugitive.vim#L695-L705) is the main thing that needs to be updated. I'm not sure how many other places need to be updated. I would start by changing `resolve(dir)` to `dir` in `FugitiveExtractGitDir()`, changing `dir` to `resolve(dir)` in that linked code, and seeing how badly it breaks. Reach out if you have further questions.

brookhong:
Thanks.

Unfortunately, `git config core.worktree $(pwd)` does not make it work. And I followed your suggestion to make change in `fugitive#Find`, it could make `Gvdiffsplit` work, but for `Gstatus`.

tpope:
I don't know what to tell you; I just tested a symlinked `.git` repo with `core.worktree` set and it worked perfectly. Does `git config --get core.worktree` return the repository root?

PR #1275, opened 2019-07-06
s:BuildShell: fix E706

blueyed:
This happens with old Vim versions (before 7.4.1546):

> Vim(let):E706: Variable type mismatch for: cmd

Ref: https://github.com/vim/vim/commit/f6f32c38b

Not sure if you want to take this, but could not find a minimal required Vim version.
I've noticed this with Neomake's test suite, where integration with vim-fugitive is tested.


tpope:
I do want to fix stuff like this, thanks!

Issue #1287, opened 2019-07-13
[Bug] Gdiff Exception not caught


skanehira:
When I run `:Gdiff` then occur the `Exception not caught` error.
I think we should catch the exception.

![image](https://user-images.githubusercontent.com/7888591/61166977-7e941700-a572-11e9-9776-5dc08955793e.png)

- error message
```
Error detected while processing function <SNR>74_Diff[55]..<SNR>74_IsConflicted[1]..<SNR>74_TreeChomp:
line    7:
E605: Exception not caught: fugitive: error running `git -C /Users/skanehira/dev/go/src/github.com/skanehira/docker.vim --literal-pathspecs ls-files --unmerged -- ''`: fatal: empty string is not a valid pathspec. please use . instead if y
ou meant to match all paths
Error detected while processing function <SNR>74_Diff:
line   55:
E171: Missing :endif
```

PR #1288, opened 2019-07-13
Fix :Gdiff Exception not caught #1287

skanehira:
relate #1288


tpope:
Thanks. I think it's better, in this case, not to throw the exception to begin with.

(Also, for future reference, `false` isn't a thing, it should be `return 0`.)

Issue #1231, opened 2019-03-11
Feature request: try to restore the cursor position when :Gedit swaps the buffers


bam80:

tpope:
Try how?

bam80:
Well, at least when the buffers are identical, it's possible to restore position precisely.
Of course, it only makes sense when no position was saved before for that buffer.

tpope:
I've warmed up to this for the zero argument version of `:Gedit`. I think the "how" is answered by `git diff` plus counting offsets from hunks.

odnoletkov:
As a workaround I currently use `:Gdiffsplit` instead of `:Gedit` (with same arguments) when I want to preserve the cursor position. This opens target buffer in a split and – thanks to Vim built-in `diffsplit` logic I believe – the cursor position is preserved. Then I just <kbd>CTRL-W_o</kbd> and I'm good to go.

PR #1326, opened 2019-08-30
Preserve cursor position when editing different version of the same path

odnoletkov:
Closes https://github.com/tpope/vim-fugitive/issues/1231

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -3893,7 +3893,7 @@ function! s:UsableWin(nr) abort
         \ index(['nofile','help','quickfix'], getbufvar(winbufnr(a:nr), '&buftype')) < 0
 endfunction
 
-function! s:OpenParse(args) abort
+function! s:OpenParse(args, can_cmd) abort
   let pre = []
   let args = copy(a:args)
   while !empty(args) && args[0] =~# '^+'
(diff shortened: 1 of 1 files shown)


tpope:
For starters, I said "zero argument version" only. `:Gedit commit:filename` should not change behavior just because you happen to be editing `filename`. I would accept extending it to the (currently undocumented) `>commit` and `>~2` notations.

Consider altering `s:OpenParse()` to return the line number as a default for the second parameter.

autoload/fugitive.vim
```diff
@@ -3913,7 +3913,38 @@ function! s:OpenParse(args) abort
   else
     let file = '%'
```
tpope: You probably want to change these defaults to `>:0` and `>` so the zero argument version also works.

autoload/fugitive.vim
```diff
@@ -3913,7 +3913,38 @@ function! s:OpenParse(args) abort
...
+    if !exec_error
+      call filter(res, 'v:val =~# "^@@"')
+      call map(res, 'substitute(v:val, ''[-+]\d\+\zs '', ",1 ", "g")')
+      call map(res, 'matchlist(v:val, ''^@@ -\(\d\+\),\(\d\+\) +\(\d\+\),\(\d\+\) @@'')[1:4]')
+      call filter(res, 'v:val[0] < '.line('.'))
+      let hunk = get(res, -1, [0,0,0,0])
+      if hunk[0] + hunk[1] > line('.')
+        let pre = ['+'.(hunk[2] + max([1 - hunk[3], 0]))]
+      else
+        let pre = ['+'.(hunk[2] + max([hunk[3], 1]) + line('.') - hunk[0] - max([hunk[1], 1]))]
```
tpope: If the user calls `:Gedit +3 >~`, it should jump to line 3, not to the found line. So the existing `pre` needs to take precedence.

Also `:Gread >~` needs to continue working. Maybe `s:OpenParse()` wasn't the best place for this after all.

autoload/fugitive.vim
tpope: I'll do this on master.

autoload/fugitive.vim
odnoletkov: Looks like +cmd doesn't currently work for `:Gread` at all:
> E484: Can't open file +3 /pat/to/file

Is it used for anything currently or just added for symmetry in https://github.com/tpope/vim-fugitive/commit/6c89a3675b1f9baf445e3a1b90fff3f558c85e31? If not maybe just remove +cmd handling from `:Gread`? It doesn't make much sense to me especially with `:{range}Gread`.

tpope: That sounds good, but we should preserve `++opt` handling. Perhaps `OpenParse()` should take a third parameter that specifies whether `+cmd` support is wanted.

autoload/fugitive.vim
odnoletkov: So do you still want to preserve cursor in plain `:Ge`? I thought of simple logic: preserve cursor iff `>…` is used.

tpope: If the filename is explicitly given (including `%`), we should edit it normally. If there is an *implicit* reference to the current file - which includes `>` and no argument - then other implicit current file behaviors are okay.

qstrahl:
@odnoletkov Hands-down my number one most wanted Fugitive feature for as long as I can remember. Thank you so much for taking the initiative! ❤️

tpope:
Ended up heavily retooling this. Thanks for the reference implementation. Let me know if you spot any regressions.

tpope:
The big changes I ended up making:

* Support merge conflicts.
* Use `diff commit1:file commit2:file` and similar incantations rather than `diff commit1 commit2 -- file`.  This simplifies conflict handling (which previously needed `MERGE_HEAD`, `CHERRY_PICK_HEAD`, etc.) and opens the door for rename support in the future.
* Avoid the use `s:PathInfo()` which won't work for work tree buffers and [can be disabled](https://github.com/tpope/vim-fugitive/blob/4f0905e963c22e633138203a7f4bb927bb9f1404/autoload/fugitive.vim#L1171-L1173) as part of a public API.
* Avoid diffing in no-op cases, like `:Gedit >:0` when always in the index. This is a minor performance concern and also fails before the initial commit.
* Drop `fugitive#Real()` since `FugitivePrepare()` already does that.

odnoletkov:
Looks great, I've tested all blob/index/work tree combinations and merge conflicts – found no issues

PR #1327, opened 2019-08-30
blame: only handle current blame buffer/window

blueyed:
This keeps 'scrollbind' etc for when blaming a buffer used with normal
diffing already etc.

Without this scrollbind etc would be unset after e.g. `:Gdiff HEAD~10:% | Gblame | q`.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -4971,21 +4971,17 @@ function! s:BlameSubcommand(line1, count, range, bang, mods, args) abort
           silent tabedit %
         endif
         let mods = substitute(a:mods, '\<tab\>', '', 'g')
+
+        let bufnr = bufnr('')
+        " Delete any existing blame buffer(s) for the current buffer.
         for winnr in range(winnr('$'),1,-1)
-          if getwinvar(winnr, '&scrollbind')
-            call setwinvar(winnr, '&scrollbind', 0)
-          endif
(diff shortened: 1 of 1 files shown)


blueyed:
Please let me know if I should fix the commit message, or if other changes are necessary, of course.

tpope:
If we're going to change this, I'd rather limit the change to when the *current* buffer has `'scrollbind'`. If there's an unrelated set of scrollbound buffers in the current tab, we should continue to disable them.

Would it make sense to disable folds in the other scrollbound buffers?

And yes, fix the commit message.

tpope:
> Would it make sense to disable folds in the other scrollbound buffers?

Attempting to answer this question turned into implementing it.

blueyed:
Thanks for addressing this, and sorry for the late followup.

My branch/PR behaved differently fundamentally, since it allowed to blame two windows simultaneously (`:Gdiff origin/master | windo Gblame`).
Is this something you would like to support?
I find this useful to see details (i.e. the git-blame) when comparing to another branch.

Without looking at the code I assume that's what you've meant with:

> I'd rather limit the change to when the _current_ buffer has `'scrollbind'`

tpope:
I just replaced `'cursorbind'` with a custom autocommand. Eliminating that from the mix makes things a bit easier to reason about, so I pushed an experimental change to allow simultaneous blames in diffed buffers.

PR #1332, opened 2019-09-05
Reveal folds when using jump maps in commits

odnoletkov:


autoload/fugitive.vim
```diff
@@ -2731,6 +2736,7 @@ function! s:PreviousItem(count) abort
     endif
   endfor
   call s:StageReveal()
+  exe 'normal! zv'
```
tpope: Looks like it'd be less repetitive to do this inside `s:StageReveal()`?

odnoletkov:
Apparently this also resolves the issue with jump maps sometimes hanging in commits. Conditions in the scrolling loop in `StageReveal` didn't always work properly with folds.

autoload/fugitive.vim
```diff
@@ -2629,6 +2629,7 @@ function! s:Do(action, visual) abort
 endfunction
 
 function! s:StageReveal(...) abort
+  exe 'normal! zv'
   let begin = a:0 ? a:1 : line('.')
```
tpope: "Open fold on a current line" sounds like the wrong behavior when acting on a line number argument.

odnoletkov: I don't see the line number arg version used anywhere. What are the sample use cases for it? I'm trying to understand if the arg version should even reveal the fold.

Also, is it even possible to reveal the folded line without jumping to it first?

tpope: Oh, perfect. Go ahead and rip out the argument entirely if you don't mind and this is good to go.

PR #1337, opened 2019-09-13
Add --porcelain=v2 status and improve submodules

odnoletkov:


tpope:
Wow, this is quite a big PR. Thanks for putting in the time.

I really, really don't like the lowercase `m`, both aesthetically and for what it does to the implementation. I don't suppose you have an idea of what to use instead? If all else fails I would double it up on `?`, as the `?`/`m` distinction doesn't seem *that* important.

odnoletkov:
Just to confirm – did you know that this format is taken straight from `git status --short`?

I'm used to small `m` for submodules. Agree that the `m`/`?` distinction is far less important than `M`/`m`. So I can definitely live with `?` for both modified and untracked submodules content.

But I think I would be pretty surprised finding first time that `?` for submodule means modified content. Especially since rest of the signs match `git status --short` exactly.

tpope:
Yes, I know the `--short` origin, and do hate to deviate from it, but our buffer has a lot more going on than `--short`, and I value the uniformity of uppercase more there.

odnoletkov:
Updated to use '?' for submodules with modified and untracked content

tpope:
In an effort to make this change more focused and easier to comprehend, I've ported over your `b:fugitive_status` nesting change and renamed to `b:fugitive_files` in the process. Can you rebase?

I believe your first 2 (3?) commits can be combined into one, as having one intermediate commit with v1 porcelain support ripped out isn't very useful.

odnoletkov:
Done.

autoload/fugitive.vim
```diff
@@ -2519,6 +2566,7 @@ function! s:StageInfo(...) abort
         \ 'paths': map(reverse(split(text, ' -> ')), 's:Tree() . "/" . v:val'),
         \ 'commit': matchstr(getline(lnum), '^\%(\%(\x\x\x\)\@!\l\+\s\+\)\=\zs[0-9a-f]\{4,\}\ze '),
         \ 'status': matchstr(getline(lnum), '^[A-Z?]\ze \|^\%(\x\x\x\)\@!\l\+\ze [0-9a-f]'),
+        \ 'sub': get(get(get(b:fugitive_files, section, {}), text, {}), 'sub'),
```
tpope: Add a `''` default to the outer `get()` here so it doesn't return `0`.

autoload/fugitive.vim
```diff
@@ -1650,48 +1650,95 @@ function! fugitive#BufReadStatus() abort
...
+
+      while i < len(output)
+        let line = output[i]
+        if line[0] ==# '?'
+          call add(untracked, {'type': 'File', 'status': line[0], 'filename': line[2:-1]})
+        elseif line[0] !=# '#'
+          if line[0] ==# 'u'
+            let file = line[161:-1]
+          else
+            let file = line[113:-1]
```
tpope: Let's use `matchstr()` for these so the long term SHA-256 migration doesn't break it.

autoload/fugitive.vim
```diff
@@ -1617,48 +1617,95 @@ function! fugitive#BufReadStatus() abort
...
-      if line[0] !~# '[ ?!#]'
-        call add(staged, {'type': 'File', 'status': line[0], 'filename': files})
+
+      while i < len(output)
+        let line = output[i]
+        if line[0] ==# '?'
+          call add(untracked, {'type': 'File', 'status': line[0], 'filename': line[2:-1]})
+        elseif line[0] !=# '#'
+          if line[0] ==# 'u'
+            let file = matchstr(line, '^ \x\{40,\} \x\{40,\} \x\{40,\} \zs.*$', 37)
```
tpope: I think you forgot part of the regex, `'^ '` won't match if `line[0] ==# 'u'`

odnoletkov: I use `matchstr(str, pat, start)` and in that case `^` seems to be matching "rest of the string". Should I just change it to plain:
```vim
matchstr(line, '^.\{37\} \x\{40,\} \x\{40,\} \x\{40,\} \zs.*$')
```
for clarity?

odnoletkov:
I'm also not 100% sure on the "Open submodule's status from :Gstatus" change.

Submodule's status is definitely what I want to jump to more often than it's directory. But this breaks symmetry with plain files which just edit the path.

If there was an alternative way to 'open' path from status, I would use it for opening status and leave the default jump action to open directory.

Another option is to change `gf` to open directory (currently it opens status).

With this version when I want to go to directory I <kbd>.e\<CR\></kbd> which is not that bad but kinda non obvious.

autoload/fugitive.vim
tpope: Uh yeah let's do that.

tpope:
Jumping to status is a little weird with the way `w:fugitive_status` is used. Let's jump to directory for now and figure out status jumping later.

odnoletkov:
Removed the 'jump to submodule status' change

tpope:
🤞

PR #1376, opened 2019-10-13
Add built-in package management installation way to the README 

pesarkhobeee:
Dear @tpope 
I am a new user in vim then I found your awesome plugins for that and I really appreciate your efforts 
The only problem was about installing them inside the new version which I knew we don't have to use an external package manager, therefore, I just added the missing part in order to make it easier to install for the people like myself :)

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -69,13 +69,12 @@ For more information, see `:help fugitive`.
 
 ## Installation
 
-If you don't have a preferred installation method, one option is to install
-[pathogen.vim](https://github.com/tpope/vim-pathogen), and then copy
-and paste:
-
-    cd ~/.vim/bundle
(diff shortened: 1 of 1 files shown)


tpope:
I have my own boilerplate for this which you can find in commentary.vim's README among other places. Feel free to port that over. :slightly_smiling_face:

pesarkhobeee:
@tpope I think right now it is ready :)

tpope:
The shell commands are good. I want to keep parity with the explanatory text as well.

pesarkhobeee:
@tpope done

pesarkhobeee:
@tpope huum I was thinking maybe you forgot this pull request! :)

PR #1381, opened 2019-10-19
Reset scrollbind/cursorbind after reblaming

osandov:
Hello, I've been hitting a bug where my buffers will sometimes end up with scrollbind/cursorbind after exiting Gblame. I finally narrowed it down to happening when I do a reblame (e.g., with ~) and figured out the root cause. See the commit message for details. Let me know if this is the proper way to fix this issue.


tpope:
Thanks! I definitely did it this way for a reason, but this code dates back all the way to the first commit. I guess we'll have to see what if anything breaks.

osandov:
Thanks for merging this! I think there's still another bug where scrollbind and cursorbind stay on when you blame a window other than the top window. I'll try to chase that down when I have some time.

PR #1384, opened 2019-10-23
Fix slow scroll and updates

yanncoupin:
Hi,

Every time I worked using remote filesystems I found vim slow or veeery slow. Today I realized this was caused by my fugitive status line. So I decided to add some cache. It's short, only 10 seconds, but it makes a huuuge difference! Scrolling is now butter smooth again and vim is much snappier in general.

Please critique if I made some stupid mistakes or merge if you like it.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -5913,17 +5913,21 @@ endfunction
 " Section: Statusline
 
 function! fugitive#Statusline(...) abort
+  if exists("w:update") && reltimefloat(reltime(w:update))<10
+      return w:status
+  endif
   let dir = s:Dir(bufnr(''))
   if empty(dir)
     return ''
   endif
-  let status = ''
+  let w:status = ''
   let commit = s:DirCommitFile(@%)[1]
   if len(commit)
-    let status .= ':' . commit[0:6]
+    let w:status .= ':' . commit[0:6]
   endif
-  let status .= '('.FugitiveHead(7, dir).')'
-  return '[Git'.status.']'
+  let w:status .= '('.FugitiveHead(7, dir).')'
+  let w:update = reltime()
+  return '[Git'.w:status.']'
 endfunction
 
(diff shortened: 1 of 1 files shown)


tpope:
A ten second cache means the statusline could be wrong for up to ten seconds. I'd rather start by making `fugitive#Head()` cached based on the value of `getftime(fugitive#Find('.git/HEAD', dir))`, as that has zero downside. If that's still too slow, then we can discuss some sort of timeout solution.

I'd also consider an option to disable Fugitive entirely for certain directories, if that's a thing that interests people.

yanncoupin:
I tried using `getftime` like you suggested (see [this commit](https://github.com/tpope/vim-fugitive/commit/6259ab63c16aa7ede5d8c4d8a8bdb6b27e976ba8) ) but it still pings the filesystem and provides no gain compared to no cache at all. I know this code doesn't use the dir parameter in the cache key, so it's not prod safe but it's enough to test the performances.

On the other hand, if you think 10 seconds is unacceptable I tried to lower that value. 1 second is a bit low but 2 seconds works great as far as perf goes and I can't imagine that a 1s average lag to update is so bad compared to keeping the interface responsive.

tpope:
You could cut the time roughly in half if you use the `getftime()` to also replace the `filereadable()` call. One operation rather than two.

For a simpler win, you could also drop the `filereadable()` call entirely and instead rescue the exception from `readfile()`.

yanncoupin:
I made additional tests and the `exists` function is costly when using sshfs. It's synchronous whereas the `mtime` function is cached. If I only use `mtime` + a `try/catch` as you suggested, the cache works just as well as the one in my initial suggestion. You can check my second experiment's code there: https://github.com/yanncoupin/vim-fugitive/tree/headcache

It's probably specific to that filesystem but in my case both solutions work just as well to help me keep my sanity. I still think the display cache is more efficient and generic, but it's your project, so chose the one you prefer.

Issue #912, opened 2017-05-11
Support git notes


a-m-s:
Please add support for the "git notes" command.

Basically, "Gnotes" should do the equivalent of "git notes edit", including the in-message help, and deletion of notes entered empty.

a-m-s:
Also, the "commit" view should display the notes in the same way that "git log" does.

I think the git notes concept supersedes what's described in #133.

tpope:
I'm on board, but this is pretty low on the priority list. The first step is to make `fugitive#BufReadCmd()`, when called from a buffer with a name like `fugitive:///path/to/repo/.git//notes/123456789abcdef....`, load the notes for the given commit into the buffer. `s:ReplaceCmd()` should be able to do most of the work, the new code will mostly be a bit of parsing and flow control. Get a proof of concept patch for this and we can discuss next steps.

a-m-s:
The pull request covers display of the notes.

The note can be edited with `:G notes edit <rev>`, but of course that opens an editor in a subprocess. It would be nice to have a solution for that too.

CervEdin:
> The pull request covers display of the notes.
> 
> The note can be edited with `:G notes edit <rev>`, but of course that opens an editor in a subprocess. It would be nice to have a solution for that too.

This fits most of my needs.

Just adding that you can `<C-R><C-G>` to get the SHA.

So you can do `:G notes edit <C-R><C-G>`

tpope:
> So I was wondering: What is the difference between opening a commit from blame with `o` vs using `:Git show` on that commit?

It's basically doing `git show` with a very specific `--pretty` format: https://github.com/tpope/vim-fugitive/blob/d4877e54cef67f5af4f950935b1ade19ed6b7370/autoload/fugitive.vim#L3224

I have long wanted to replace this with something better, but it's tricky, because a lot is tightly coupled to the exact format of this output, most notably the "jump to file under cursor" code that's a big ball of mud. And I don't want something that's _too_ complicated—running several different commands, processing their output, and stitching it together—for performance and maintenance reasons.

PR #1406, opened 2019-11-21
Include git notes in commit buffers

a-m-s:
Issue #912

Calling "git notes list" serves both to suppress the "Notes:" prefix when a
commit does not have any notes, and to avoid using notes formats in older
versions of git.


a-m-s:
I updated the commit message. Is it acceptable now?

tpope:
The commit message failure is because I require the body to be wrapped to 72 characters, not 80.

My hesitation, though, is that I've bent over backwards to keep this to a single `git` execution for performance reasons (especially on win32 where shelling out has a massive overhead), and it's hard to justify tossing that out the window for a niche feature like notes. I'd be more amenable if it did always used the notes format and then deleted the `Notes` header after the fact.

But I also don't want to encourage you to spend more time on this because I plan to scrap the current implementation entirely and replace it with something closer to the new `:Gstatus`. This would likely manifest itself by parsing the output of `git show` rather than blindly shoving it into the buffer, which would make removing that `Notes` header downright trivial. This change is pretty high up on my todo list, but it's not quite at the top.

a-m-s:
Do you happen to know what an old git would do with the `%N`? If it's harmless then I can figure that out, I think.

a-m-s:
The patch is now updated. As is the commit message.

I imagine you know a better way to do the text editing; my vim-script experience is fairly limited. This works though. There's no `Notes:` prefix when no notes exist. Nor when notes are not supported.

I tested that `%Z` is just passed through, so I'm presuming that old git would do the same with `%N`.

tpope:
You can use `fugitive#GitVersion(1, 7)` to check for a Git version with [`%N` support](https://github.com/git/git/commit/8b208f0213ed349ecb2ab8f7bb6c5072f8011a70).

a-m-s:
> You can use `fugitive#GitVersion(1, 7)` to check for a Git version with [`%N` support](https://github.com/git/git/commit/8b208f0213ed349ecb2ab8f7bb6c5072f8011a70).

OK, that's been in Git about 5 years longer than I thought it might! We shouldn't need to check; the patterns should work anyway. I tested it with `%Z` (which isn't defined right now), and it did the right thing.

tpope:
I bring it up as a possible simplification because I haven't deciphered what `FuGiTiVeEnDmArKeR` is for but it's raising all my alarm bells.

a-m-s:
I needed a way to remove `Notes:` when there is no note there, without running any other git command to check. It seems likely that `Notes:` could appear in the commit message or in the commit diff itself, so a unique marker was helpful at the start. I also needed an end marker, and `^diff ...` didn't seem appropriate.

You may well know a better way.

PR #1417, opened 2019-12-14
fugitive://foo should not to be left instead of edited buffer

kaile256:
Any problem with this small addition?

The reason to use :windo instead of :bufdo is that the latter sometimes fails to :quit fugitive://foo, especially diff-buffers are in horizontal.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -4095,6 +4095,12 @@ function! s:OpenParse(args, wants_cmd) abort
 endfunction
 
 function! s:DiffClose() abort
+  for bufnr in tabpagebuflist()
+    if bufname(bufnr) =~# '^fugitive://'
+      execute bufwinnr(bufnr) .'windo quit'
+    endif
+  endfor
+
   let mywinnr = winnr()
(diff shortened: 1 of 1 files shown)


tpope:
You can't close all `fugitive://` buffers. I might be diffing between two different `fugitive://` buffers. I might have a third `fugitive://` buffer open that isn't being diffed at all.

Please start by explaining what problem you're experiencing and how to reproduce it.

kaile256:
I'm sorry for the naive opinion.  I didn't expect to compare only previous commits when I made this change. 
All I wanted was not to leave `fugitive://` buffers which was created by `:Gdiffsplit` on a normal buffer. Does no professional git user usually expect to leave a normal buffer, instead of `fugitive://`, in such a case.

tpope:
I can't help you if you won't tell me how to reproduce the problem.

kaile256:
The case is `:Gdiffsplit` in a normal buffer and then `dq` in a buffer created by `:Gstatus`.

But this change turned out incomplete, I drop this pr.

PR #1433, opened 2020-01-15
Use jobs in lieu of `:!` for :Git

tpope:
This solves several problems, big and small, with using `:!`:

* Inconsistent behavior across platforms.  `:!` uses the current terminal in
  terminal Vim, a dumb pseudo-TTY in GVim, and a pipe in Neovim.
* If Git needs a text editor, it spawns it as a child process, which is an
  unsatisfying experience using terminal Vim and broken on GVim and Neovim.
  This has necessitated the creation of command wrappers like `:Gcommit` and
  `:Grebase`, and these wrappers are imperfect, leaky abstractions.  Many
  commands, such as `git tag`, need wrappers but lack them.
* The pager experience is similarly flawed.  We currently work around this
  with `--no-pager` on GVim and Neovim, and by intercepting particularly
  chatty commands like `git log` and loading them into a temp buffer instead.
* Neovim's use of a pipe makes it impossible for commands like `git push` to
  ask for a password.  Earlier versions would hard lock if a command attempted
  to do so.
* A press enter prompt is always issued, even for low ceremony commands like
  `:Git add %`.  This not only directly inconveniences the user, but also
  prevents the use of `:Git` in anywhere in Fugitive's internals that we don't
  want a prompt, such as the `s`, `u`, and `-` maps in `:Gstatus`.

This pull request uses Vim 8's `job_start()` and Neovim's `jobstart()` for the
default `:Git` executor.  The `pty` flag is passed to get better handling of
passwords, among other things, and `TERM` is forced to `dumb` to disable ANSI
escape sequences.  I have no plans to support ANSI colors, but I'll entertain
optionally using an external library to implement them if one is available.

Fugitive has been updated to take advantage of the new job executor in the
following ways:

* Custom `:Grebase`, `:Gmerge`, and `:Gpull` implementations completely
  removed.  Using them on Vim 7 will fall back to `:!`, which seems good
  enough for the outdated server use case.  This drops the old behavior of
  loading the output into the quickfix list, which I always found to be more
  annoying than helpful.  For loading conflicts into the quickfix list, use
  the new native `:Git mergetool`.
* `:Gcommit` and `:Grevert` use the job executor if available, and fall back
  to the custom implementation if not.  (I'll probably end up ripping out this
  custom implementation as well.)  This means the old "`:Gcommit -v` in new
  tab" behavior is gone.  It could be brought back, but since it's also
  available on `:tab Gcommit -v`, I've moved it to the `cv` maps instead.
* `:Gpush` and `:Gfetch` now only use their `:make`/`:Make` based quickfix
  capture if a bang (`!`) is passed.  This probably won't make it into the
  release, but it will enable people to test that password input is working as
  expected.

**I need people to test the `job` branch**, especially users of Neovim and
*especially* users of Windows.  Please comment with any issues.  Long
term testing is especially helpful; I want to see what corner cases pop up.
I aim to merge this after the next Fugitive release, hopefully within a month
or two.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -3603,109 +3603,28 @@ function! s:CommitInteractive(line1, line2, range, bang, mods, args, patch) abor
   endif
 endfunction
 
-function! s:CommitSubcommand(line1, line2, range, bang, mods, args, ...) abort
-  let mods = substitute(s:Mods(a:mods), '\C\<tab\>', '-tab', 'g')
-  let dir = a:0 ? a:1 : s:Dir()
-  let tree = s:Tree(dir)
-  let msgfile = fugitive#Find('.git/COMMIT_EDITMSG', dir)
-  let outfile = tempname()
-  try
-    if s:winshell() || &shellcmdflag ==# '-Command'
-      let command = 'set GIT_EDITOR=false& '
+function! s:CommitSubcommand(line1, line2, range, bang, mods, args) abort
+  let argv = copy(a:args)
+  let i = 0
+  while get(argv, i, '--') !=# '--'
+    if argv[i] =~# '^-[apzsneiovq].'
+      call insert(argv, argv[i][0:1])
+      let argv[i+1] = '-' . argv[i+1][2:-1]
     else
-      let command = 'env GIT_EDITOR=false '
+      let i += 1
     endif
-    let argv = a:args
-    let i = 0
-    while get(argv, i, '--') !=# '--'
-      if argv[i] =~# '^-[apzsneiovq].'
-        call insert(argv, argv[i][0:1])
-        let argv[i+1] = '-' . argv[i+1][2:-1]
-      else
-        let i += 1
-      endif
-    endwhile
-    let command .= s:UserCommand(dir, ['commit'] + argv)
-    if (&autowrite || &autowriteall) && !a:0
-      silent! wall
-    endif
-    if s:HasOpt(argv, '-i', '--interactive')
-      return s:CommitInteractive(a:line1, a:line2, a:range, a:bang, a:mods, argv, 0)
-    elseif s:HasOpt(argv, '-p', '--patch')
-      return s:CommitInteractive(a:line1, a:line2, a:range, a:bang, a:mods, argv, 1)
-    else
-      let [error_string, exec_error] = s:TempCmd(outfile, command)
-      let errors = split(error_string, "\n")
-    endif
-    if !has('gui_running')
-      redraw!
-    endif
-    if !exec_error
-      echo join(errors, "\n")
-      if filereadable(outfile)
-        echo join(readfile(outfile), "\n")
-      endif
-      call fugitive#ReloadStatus(dir, 1)
-      return ''
-    else
-      let error = get(errors,-2,get(errors,-1,'!'))
-      if error =~# 'false''\=\.$'
-        let i = 0
-        while get(argv, i, '--') !=# '--'
-          if argv[i] =~# '^\%(-[eips]\|-[CcFm].\+\|--edit\|--interactive\|--patch\|--signoff\|--reedit-message=.*\|--reuse-message=.*\|--file=.*\|--message=.*\)$'
-            call remove(argv, i)
-          elseif argv[i] =~# '^\%(-[CcFm]\|--reedit-message\|--reuse-message\|--file\|--message\)$'
-            call remove(argv, i, i + 1)
-          else
-            if argv[i] =~# '^--cleanup\>'
-              let cleanup = 1
-            endif
-            let i += 1
-          endif
-        endwhile
-        call insert(argv, '--no-signoff', i)
-        call insert(argv, '--no-interactive', i)
-        call insert(argv, '--no-edit', i)
-        if !exists('cleanup')
-          call insert(argv, '--cleanup=strip')
-        endif
-        call extend(argv, ['-F', msgfile], 'keep')
-        if (bufname('%') == '' && line('$') == 1 && getline(1) == '' && !&modified) || a:line2 == 0
-          execute mods . 'keepalt edit' s:fnameescape(msgfile)
-        elseif s:HasOpt(argv, '-v') || mods =~# '\<tab\>'
-          execute mods . 'keepalt -tabedit' s:fnameescape(msgfile)
-        else
-          execute mods . 'keepalt split' s:fnameescape(msgfile)
-        endif
-        let b:fugitive_commit_arguments = argv
-        setlocal bufhidden=wipe filetype=gitcommit
-        return '1'
-      elseif empty(errors)
-        let out = readfile(outfile)
-        echo get(out, -1, '') =~# 'stash\|\d' ? get(out, -2, '') : get(out, -1, '')
-        return ''
-      else
-        echo join(errors, "\n")
-        return ''
-      endif
-    endif
-  catch /^fugitive:/
-    return 'echoerr ' . string(v:exception)
-  finally
-    call delete(outfile)
-  endtry
+  endwhile
+  if s:HasOpt(argv, '-i', '--interactive')
+    return s:CommitInteractive(a:line1, a:line2, a:range, a:bang, a:mods, argv, 0)
+  elseif s:HasOpt(argv, '-p', '--patch')
+    return s:CommitInteractive(a:line1, a:line2, a:range, a:bang, a:mods, argv, 1)
+  else
+    return {}
+  endif
 endfunction
 
 function! s:RevertSubcommand(line1, line2, range, bang, mods, args) abort
-  let dir = s:Dir()
-  let no_commit = s:HasOpt(a:args, '-n', '--no-commit', '--no-edit', '--abort', '--continue', '--quit')
-  let cmd = s:UserCommand(dir, ['revert'] + (no_commit ? [] : ['-n']) + a:args)
-  let [out, exec_error] = s:SystemError(cmd)
-  call fugitive#ReloadStatus(dir, 1)
-  if no_commit || exec_error
-    return 'echo ' . string(substitute(out, "\n$", '', ''))
-  endif
-  return s:CommitSubcommand(a:line1, a:line2, a:range, a:bang, a:mods, [], dir)
+  return {'args': ['revert', '--edit'] + a:args}
 endfunction
 
 function! fugitive#CommitComplete(A, L, P) abort
@@ -3727,16 +3646,6 @@ endfunction
 
 function! fugitive#RevertComplete(A, L, P) abort
   return s:CompleteSub('revert', a:A, a:L, a:P, function('s:CompleteRevision'))
-endfunction
-
-function! s:FinishCommit() abort
-  let buf = +expand('<abuf>')
-  let args = getbufvar(buf, 'fugitive_commit_arguments')
-  if !empty(args)
-    call setbufvar(buf, 'fugitive_commit_arguments', [])
-    return s:CommitSubcommand(-1, -1, 0, 0, '', args, s:Dir(buf))
-  endif
-  return ''
 endfunction
 
 " Section: :Gmerge, :Grebase, :Gpull
@@ -4618,11 +4527,6 @@ function! fugitive#WqCommand(...) abort
     return result.'|quit'.bang
   endif
 endfunction
-
-augroup fugitive_commit
-  autocmd!
-  autocmd VimLeavePre,BufDelete COMMIT_EDITMSG execute substitute(s:FinishCommit(), '\C^echoerr \(''[^'']*''\)*', 'redraw|echohl ErrorMsg|echo \1|echohl NONE', '')
-augroup END
 
 " Section: :Gpush, :Gfetch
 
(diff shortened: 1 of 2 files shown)


tpope:
Windows users in particular, please comment if it works. I don't have a convenient way to test there.

fleischie:
Hello. I tried to test this branch as by your request.

Adding, committing (even special characters like spaces, etc.), rebasing all work ok. Although the output all had weird ASCII color glyphs (I assume).

The only issues I had were:

Merging:
Annotation 2020-01-15 192916

Reverting:
Annotation 2020-01-15 193052

I was running `neovim` in `cmd.exe`, which I installed with Chocolatey (if that is important). If you need further info, let me know, although I have to admit, that I am not too well versed in the forbidden talents in using Windows.

:v:

tpope:
I'm not sure what to do about the escape codes. They went away on Linux Neovim when I changed `TERM` from `ansi` to `dumb`. (In plain Vim the default was already `dumb`.) Hopefully there's a Windows equivalent but I'm not sure what it would be.

That first issue you cite was a simple bug on my part. That second one is confusing. `:Gcommit` works but `:Grevert` fails? I'm not sure what could cause that. The double layer of quoting - `'"sh C:\...\1.sh"'` - makes me wonder if it's treating the leading `sh` as part of the file name. That sounds like a Git bug if so. What Git version are you using and how was it installed?

I'm pushing a change so that you can call `:echo g:fugitive_last_job.cmd` after a command runs to see the exact command that was invoked. If you could try that revert again and see what the command was, that would be awesome.

fleischie:
I will have to check in approx. 10-12 hours because of timezones and sleep.

I just noticed I never tried to just run `:Gcommit`, I always supplied an `-m` flag. 🤦

Tomorrow I can check on both `cmd.exe` and `PowerShell`, what git version I use, what `:Gcommit` w/o a message/file does, and report back.

Also reading the above warning in detail: I wonder whether Windows knows how to handle `sh` on the command line, but I'll check tomorrow.

tpope:
> I will have to check in approx. 10-12 hours because of timezones and sleep.
> 
> I just noticed I never tried to just run `:Gcommit`, I always supplied an `-m` flag. 🤦

That's actually a relief, at least it's consistently failing.

> Tomorrow I can check on both `cmd.exe` and `PowerShell`, what git version I use, what `:Gcommit` w/o a message/file does, and report back.
> 
> Also reading the above warning in detail: I wonder whether Windows knows how to handle `sh` on the command line, but I'll check tomorrow.

Running from inside `git` should include `sh.exe` in the path.

I'm pretty sure what's happening is we're doing `set GIT_EDITOR="sh C:\..."`, and those quotes are making it into the environment variable because cmd.exe is a trash fire. You could try removing the `s:shellesc` from [this line](https://github.com/tpope/vim-fugitive/blob/08601f221a336d13a606b33714e5a1e8457b07f7/autoload/fugitive.vim#L453) to test that.

fleischie:
Good morning. As promised I checked out the newest version of your work and now I don't seem to have anything working anymore. 😭  (Git version is: "2.25.0.windows.1") Disclaimer: It is a different machine than yesterday, so maybe there are other problems working here.

With `set shell=cmd.exe` it just opens a terminal prompt, but doesn't actually do anything:

Annotation 2020-01-16 085900

With `set shell=powershell` it doesn't even show the terminal prompt. Additionally it still completely fails to do anything regarding temporary files:

Annotation 2020-01-16 085635

tpope:
The fact you're seeing `:!` makes me think you're on the master branch, or you have a very old Neovim. Does `:echo exists('*jobstart')` report `1`?

fleischie:
It does in fact report `1`.
The curious thing with my setup (neovim v0.4.3 on powershell) is: Running `system('uname')` returns `uname` (i.e. the input string, same as with other commands), and I noticed the help section for `system` mentions it does work similar to `jobstart`. `v:shell_error` also produces `0`.

Same on `cmd.exe`... 🤔

tpope:
You're sure you're on the job branch? Other weirdness aside, that code should never run on the job branch if `jobstart()` exists.

Try with both `:Git status` and `:Git commit`.

fleischie:
You are absolutely right, my vim package manager is borked and didn't switch to the correct branch.

fleischie:
Now it shortly displays some control char stuff with an "OK", and then this without doing anything to/with the file:

Annotation 2020-01-16 110523

Also `git log` yields:

Annotation 2020-01-16 110619

tpope:
Try `:echo "line1\r\nline2"`. What does that show. It *should* show

```
line1
line2
```

But I have a hunch it will show

```
<blank line>
line2
```

Try `:let g:fugitive_pty = 0` and then try testing again. Based on what you've shared, I expect it will work, other than things that require a TTY like password entry.

The `git log` issue is definitely related to your "troubles calling functions that operate on temporary files" thing you cited earlier. Something in your setup is broken, and I'm not sure what it is. Appears not to affect job testing so *I* am not worried. :wink:

Thanks for your help!

fleischie:
Your hunch was right, it returned:
```
<blank line>
line2
```

Setting `:let g:fugitive_pty = 0` did completely silence the output window. But still I don't think it does anything. But on the other hand this could be related to the temp-file issue I am having. 😢 

Thank you for your effort. Is there another way you can be supported (money-wise?) (apart from Github (with whom I don't wanna do business with))?

tpope:
I think it's silent because of the `\r` issue, which appears to be a bug specific to Windows Neovim. The ANSI escape sequences being displayed appear to also be specific to Windows Neovim with a PTY job; it's less clear if this is a bug or a "feature". I pushed a couple of changes that should compensate for both by disabling the PTY on the problematic platform and stripping out carriage returns. Not really happy about it but it's better than nothing.

> Thank you for your effort. Is there another way you can be supported (money-wise?) (apart from Github (with whom I don't wanna do business with))?

GitHub matches sponsorships for the first year. If your ethics won't allow even a transaction that costs them money, there's also a Paypal link on the sponsor button. Do not feel obligated; you're giving back just by helping me ship this feature.

habamax:
Hi @tpope check, the minor issue I have so far (latest gvim8.2.139), win10:

vim-fugitive-jobs

After `:Gpush` command I can type anything and it is echoed "in command line" (whatever it is called)

Basically it blocks vim -- I cannot switch to another window, for example.

tpope:
Is it asking for a username/password? Does it clear up if you press enter a couple of times?

habamax:
> Is it asking for a username/password?

No



> Does it clear up if you press enter a couple of times?

Yes, on a second Enter it is cleared.

habamax:
The same issue with `:Gcommit` -- right after it I want to switch window and move on pressing `<C-w>wjjjll`

image

tpope:
I just pushed a commit to swallow that error. I saw it a bit when I got a chance to test on Windows the other day but couldn't pin it down.

Two enters sounds a loooooot like username/password, even if no prompts are showing. Is it that behavior specific to `:Gpush` or is it happening for `:Gcommit` too?

habamax:
> Is it that behavior specific to `:Gpush` or is it happening for `:Gcommit` too?

Gcommit too, but I will dbl check it.

tpope:
Both push and commit have associated hooks. Do you have any enabled?

habamax:
With the new commit I have the following behavior:

1. It is still blocking normal operations (can only type that has no immediate effect)
2. all typed things are applied to Git status or commit buffer (although `C-ww` has no window switch effect)

vim-fugitive-jobs3

habamax:
> Both push and commit have associated hooks. Do you have any enabled?

No.

I only have
```
if executable("git")
	silent! packadd vim-fugitive
	silent! packadd vim-flog
endif
```

tpope:
I mean Git hooks, in `.git/hooks`, or in `git config --get core.hooksPath`.

Are you having better results with dumb commands like `:Git ls-files`?

habamax:
> I mean Git hooks, in `.git/hooks`, or in `git config --get core.hooksPath`.

Oh, I see. No I don't have anything there

> Are you having better results with dumb commands like `:Git ls-files`?

well it works without issues as far as I can see (never used it with prev fugitive, don't know if it differs)

image

tpope:
Yeah that looks fine, but just to make sure try with a command that's brief enough to not need pagination (`:Git status` might work).

For `:Gcommit`, the issue might be that stdout was buffered. I saw this for some other commands, like  `:Git add -p` when I disabled the `:terminal` intercept. I didn't see it for `:Gcommit` but that could be a difference in how Git was compiled. Try adding `\n` (backlash, n) to the following line, immediately after the `\007` and before the last double quote:

https://github.com/tpope/vim-fugitive/blob/e37f21e22468199a8b9e4e2ce20a10f021828d3f/autoload/fugitive.vim#L2383

If it's *line* buffering, that will fix it, at the expense of echoing an extra newline. If it's *block* buffering, then you'll need to add, uh, 4096 spaces, or whatever the block size is.

habamax:
> Try adding `\n` (backlash, n) to the following line, immediately after the `\007` and before the last double quote:

Ok, will do it tomorrow morning.

habamax:
> For `:Gcommit`, the issue might be that stdout was buffered. I saw this for some other commands, like `:Git add -p` when I disabled the `:terminal` intercept. I didn't see it for `:Gcommit` but that could be a difference in how Git was compiled. Try adding `\n` (backlash, n) to the following line, immediately after the `\007` and before the last double quote:


Decided to do it tonight :)

I have added this:
```
          \ 'printf "\033]51;fugitive:edit\007\n"',
```
And tried `Gcommit` -- the same (or maybe not the same but feels the same):
1. it echoes chars as soon as I start typing
2. When Gcommit opens buffer -- all my typing still goes there

vim-fugitive-jobs4

habamax:
> eah that looks fine, but just to make sure try with a command that's brief enough to not need pagination (`:Git status` might work).

With `Git status` if I start typing I can see it in the output:

image

habamax:
> If it's _block_ buffering, then you'll need to add, uh, 4096 spaces, or whatever the block size is.

Where to put it? :)

tpope:
Ohhhhh I bet this is something different. I ran into an issue where `getchar()` would block even though `getchar(1)` said a character was available. That's what the second half of this conditional is about:

https://github.com/tpope/vim-fugitive/blob/e37f21e22468199a8b9e4e2ce20a10f021828d3f/autoload/fugitive.vim#L2253

Try adding `:echo peek` immediately after that line. If it immediately echoes, it sounds like I need to further amend the conditional.

You can also try changing the `getchar()` to `getchar(0)` which shouldn't block, but I believe when I tested it still blocked.

habamax:
There is also `:Grebase` after it finishes don't let me do anything till I press Enter: 

image

habamax:
> Try adding `:echo peek` immediately after that line. If it immediately echoes, it sounds like I need to further amend the conditional.

vim-fugitive-jobs5

habamax:
> You can also try changing the `getchar()` to `getchar(0)` which shouldn't block, but I believe when I tested it still blocked.

```
      let peek = getchar(1)
      if peek != 0 && !(has('win32') && peek == 128)
		  echo peek
        let c = getchar(0)
        let c = type(c) == type(0) ? nr2char(c) : c
        call s:RunSend(a:job, c)
        if !a:state.pty
          echon c
        endif
      endif
```

Didn't help.

tpope:
That `:Gpush` looks fine, and I don't see you pressing enter this time around. Are you sure it's not just slow?

I pushed a new commit that closes stdin entirely and eliminates all keypress handling. It should work perfectly or hang indefinitely. Hopefully the former.

tpope:
So it sounds like it's working correctly, and the only issue is that if you mash the keyboard while it's running, those key presses get queued up for processing by Vim, and closing stdin unsurprisingly didn't change that. I don't think that's a deal breaker; it's a pretty common behavior when Vim is doing something slow.

It is a little surprising how slow it is. One thing that could affect that is having the status window open (and automatically reloading). What version of Vim are you using? What is `'shell'`?

habamax:
> It is a little surprising how slow it is. One thing that could affect that is having the status window open (and automatically reloading). What version of Vim are you using? What is `'shell'`?

gvim 8.2.140, shell is `C:\Windows\system32\cmd.exe`

habamax:
> I don't think that's a deal breaker; it's a pretty common behavior when Vim is doing something slow.

Hmm. Indeed. With at least one exception: the old `Gpush`. While slow, it lets me continue using vim:
vim-fugitive-jobs7

tpope:
If you want, you can try things with no status window open and see if it's any faster. But at this point I think it's good enough. I'll probably merge everything but the closed stdin change to master in the next day or two, and entertain further tweaks if the masses complain. This is such a massive usability improvement that I am accelerating the schedule.

I'll keep the old `:Gpush` around too. The next step is to replace the old `:Make`/`:make` based runner it is using with a job running variant that routes to the quickfix list, and make it available to all commands. Not sure what the interface will be, perhaps `:Git!` or something clever like `:Git --no-pager`. This is nearly as big of task as this PR was, not sure when I'll knock it out.

tpope:
@fleischie I did get a chance to try the Neovim/PowerShell combo and ran into the exact same `system('uname')` behavior you did. I fixed it by copying the exact settings that GVim used (although this wasn't quite perfect, I could still break it with certain shell escaping constructs that worked fine on GVim):

```vim
set shell=PowerShell shellcmdflag=-Command shellquote= shellxquote= shellxescape=\"\|<>()@^ shellpipe=> shellredir=>
```

It's kind of bonkers that this is broken out of the box, given that the PowerShell based Chocolatey is a recommended Neovim installation method.

habamax:
> If you want, you can try things with no status window open and see if it's any faster.

It is almost the same. `Gcommit` is slightly faster.

tpope:
Thanks @fleischie and @habamax.

justinmk:
> It's kind of bonkers that this is broken out of the box, given that the PowerShell based Chocolatey is a recommended Neovim installation method.

Not for lack of effort. I think there was a big in 0.4.3 which would be fixed in 0.5.

Also `:help powershell`

PR #1459, opened 2020-02-06
Fix typo --include-index to --keep-index in help

hitsuji-no-shippo:
https://github.com/tpope/vim-fugitive/blob/0e6f72b005312225ffb47290500941d59eb1f0f7/autoload/fugitive.vim#L5858


tpope:
Thanks!

PR #1461, opened 2020-02-08
add rk and rx keymap into rebase maps of help

hitsuji-no-shippo:
https://github.com/tpope/vim-fugitive/blob/0e6f72b005312225ffb47290500941d59eb1f0f7/autoload/fugitive.vim#L5879-L5881


tpope:
These aren't official and are deliberately undocumented.

PR #1469, opened 2020-02-18
Fix Trailing Characters Invalid range error for :<count>Gstatus

woodstok:
Correct the position of newly added keepalt option from
'botright <count>keepalt split' to 'botright keepalt <count>split'


tpope:
Good catch, thanks!

PR #1470, opened 2020-02-19
Respect 'wildignorecase' when completing :Gedit

bobrippling:
This makes completion for `:Gedit` respect `wildignorecase` and `fileignorecase`, when completing the path part of a `fugitive-object`. With this change, the `:Gedit` command feel more native to vim, matching the behaviour of the `:e` command (amongst others).

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -117,12 +117,13 @@ function! s:Resolve(path) abort
   return path
 endfunction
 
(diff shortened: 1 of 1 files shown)


tpope:
Can you rename to `s:FileIgnoreCase()`? I'm trying to standardize on camel case for all new functions.

bobrippling:
Consider it done :+1:

PR #1491, opened 2020-03-11
Use S:Map() for more mappings

runiq:
There were a few mappings left that weren't customizable with
`g:[onx]remap`. This makes them configurable.


tpope:
Good catch on the help mappings, thanks!

kevinhwang91:
Maybe `g:[onx]remap` should be added a prefix like `g:fugitive_[onx]remap`.

PR #1513, opened 2020-04-14
Add help for the staging buffer mappings

attilapiros:


tpope:
Adding an `s?` map makes the `s` map hang. Adding a line that starts with `-` breaks the diff parsing.

I'm not interested in pursuing this further, but would consider a one liner like `Help: g?` line or something like that, where pressing enter on that line also opened the help.

attilapiros:
Thanks for looking into it. I totally missed the hanging `s`. 

The help lines would start with `"`:
Screenshot 2020-05-05 at 9 37 24

PR #1555, opened 2020-06-23
Fix wrong command capitalization in README.markdown

SLeitgeb:
Some command names in README.markdown have wrong capitalization, e.g. `:GMove` should be `:Gmove`. Fix to avoid confusing new users.


tpope:
The README is correct. The lowercase versions are softly deprecated. [This is documented](https://github.com/tpope/vim-fugitive/blob/27a5c3abd211c2784513dab4db082fa414ad0967/doc/fugitive.txt#L618-L622).

Issue #1563, opened 2020-07-13
Open or close all inline diffs in :G


odnoletkov:
The best way I found to do that in the current version is to `<` or `>` on both 'Staged' and 'Unstaged' section headers. This is tedious even if you use `gu` and `gs` maps.

Is there a better way to open/close all inline diffs?

If no, can I propose making `<` and `>` anywhere else (outside of sections) to do that?

nhooyr:
I think an option to have all inline diffs open by default would be nice too. Then u can just use vim's folding to open/close diffs.

nhooyr:
I think https://github.com/jreybert/vimagit may be a better fit for this anyway.

See also https://github.com/tpope/vim-fugitive/issues/569

nhooyr:
Tried it out but got weird errors when reopening vimagit sometimes about the magit window being incorrect or unmodifiable.

Writing some viml to automate this for vim-fugitive.

nhooyr:
```viml
  augroup nhooyr_fugitive
    autocmd!
    autocmd FileType fugitive normal guk>gUk>gsk>gg0zM
  augroup END
```

Will also need to use my branch from https://github.com/tpope/vim-fugitive/pull/1565 until it's merged.

nhooyr:
```viml
  augroup nhooyr_fugitive
    autocmd!
    autocmd FileType fugitive normal guk>gg
    autocmd FileType fugitive normal gUk>gg
    autocmd FileType fugitive normal gsk>gg
  augroup END
```
For some reason with them all in the same `:normal`, it doesn't expand diffs if you only have a single stage section and no unstaged/untracked. Separating them out makes things work as expected.

🤷

tpope:
"Anywhere else" is bad, but adding that behavior on the `Head:` line aka line 1 sounds reasonable. I'll implement it.

I'm also tempted to add maps like `z<` to collapse all diffs in the current section. When I want to collapse everything that's usually the core of what I'm after: I'm done with the section and want to collapse it all. But I'm not sure that justifies 3 more maps that don't really have a great mnemonic.

odnoletkov:
Cool, `<` on the `Head:` would work for me. The main use case is staging multiple files with large hunks – I often get lost after performing several actions while not seeing section headers. Folding everything down helps by showing high-level perspective.

What do you think about `<` on already closed file closes all other files in the section? That way `<<` anywhere in the section will always fold it up.

tpope:
[...]

PR #1565, opened 2020-07-20
Add ftplugin for fugitive filetype

nhooyr:
Allows users to override foldlevel as they can autocmd on FileType.

See #1563

Right now foldlevel is unconditionally set to 1 after reading the file.

An autocmd on BufWinEnter does not work either.


tpope:
The reason the code is in the current location is because it's _not_ supposed to run on every `FileType` event.  It should only run for `:Git` and not for `:edit .git/index`.

But I think I would be find with ditching the folding defaults entirely. It's not really part of the standard workflow.

PR #1566, opened 2020-07-20
Document magit like status interface in README

nhooyr:
This is a head lining feature more people should be aware of.


tpope:
See bullet 6 :slightly_smiling_face:.

PR #1585, opened 2020-08-25
Fix doc. Mark `C` mapping as deprecated in the user documentation

ademin:
Since `C` mapping has been deprecated and there is an error message on
its usage, the user documentation should have a note about it.


tpope:
`C` as a shortcut for `:Git commit` in the status buffer was deprecated ages ago and has since been removed. This `C` isn't in the status window, it's in blobs and trees. It's unrelated and very much *not* deprecated.

PR #1587, opened 2020-08-27
Add command Gdiffoff that calls diffoff and closes buffers

allopislozano:
Sometimes when you call Gdiff on a cpp file and afterwards you call Gdiff (maybe by mistake) on a different file then both diffs are mixed and the visualization is not useful.  

In this situation Gdiffoff lets the user reset the tab


tpope:
I don't know that this rises to the level of justifying a command, and if it does I'm not sure this is a good name for it. I've made the function public so you can define your own command locally.

PR #1620, opened 2020-11-09
Add short-date to Glog format

klesh:
I was building picom with latest commit, unfortunately it was unstable, but I knew it was OK couple months back.
When I tried falling back to previous build, I saw only hash and commit summary were displayed.
I think it would be helpful if short-date also displayed as well.


tpope:
This isn't up for bikeshedding. There's a configuration variable like 3 columns over if you want to change this locally.

PR #1633, opened 2020-11-25
Update README statusline help to match the one in plugin's help

rafaelleru:


tpope:
`FugitiveStatusline()` is preferred, and is in fact used in the help. `fugitive#statusline()` is a deprecated alias.

PR #1659, opened 2021-01-20
Fix bang not being passed to :Dispatch command

jceb:
I noticed that the quickfix window opened every time I called `:Git push` regardless of the presence of a bang. This pull request fixes the issue by passing the bang to `:Dispatch`.


tpope:
Finally pulled the trigger on removing this entirely in https://github.com/tpope/vim-fugitive/commit/d4bcc75ef6449c0e5592513fb1e0a42b017db9ca.

Issue #1680, opened 2021-02-20
Minimap and fugitive and list index out of range in ReloadTabStatus


ianhomer:
Minor one this (and avoidable), however others could trip over it so thought best to share ...

vim config

```
let g:vim_dir = "~/.vim"

call plug#begin(g:vim_dir."/plugged")
  Plug 'tpope/vim-fugitive'
  Plug 'wfxr/minimap.vim'
call plug#end()

let g:minimap_auto_start = 1
let g:minimap_close_filetypes = ['fugitive']
```

when I commit a change, e.g

* :Gstatus
* s (to stage)
* cc (to add message)

Then I get the error

```
"~/projects/try/try-vim/.git/COMMIT_EDITMSG" 11L, 309B
Error detected while processing function fugitive#Command[155]..<SNR>36_RunWait[46]..fugitiv
e#ReloadStatus[8]..<SNR>36_ReloadTabStatus:
line    4:
E684: List index out of range: 3
E116: Invalid arguments for function getbufvar(tabpagebuflist(tab)[winnr-1], 'fugitive_type'
) ==# 'index'
```

I think this is happening because minimap is closing one of the windows whilst the fugitive plugin is iterating through a loop, so that that the window that is the loop index does not exist anymore.

Minor issue, since I was only experimenting with minimap and error was only happening because I was trying to configure minimap to close minimap when I open fugitive.  And if I remove the following line ...

```
let g:minimap_close_filetypes = ['fugitive']
```

then all works OK for me (although I don't get quite the behaviour I was seeking).

However, perhaps other plugin interactions could cause a similar issue, so I thought I prep [PR](https://github.com/tpope/vim-fugitive/pull/1681) to fix this in case it helps.

Thanks

PR #1681, opened 2021-02-20
#1680 : Safety when window is closed during ReloadTabStatus reload loop

ianhomer:
Candidate fix for https://github.com/tpope/vim-fugitive/issues/1680

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -2884,29 +2884,26 @@ endfunction
 function! s:ReloadTabStatus(...) abort
   let mytab = tabpagenr()
   let tab = a:0 ? a:1 : mytab
-  let wins = []
-  for winnr in range(1, tabpagewinnr(tab, '$'))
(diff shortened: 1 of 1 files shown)


tpope:
This looks to me like it could still fail if you have 2 status windows open in the same tab. My preferred solution would be to change to a while loop, so that `tabpagewinnr(tab, '$') is recomputed each time. 

```vim
let winnr = 1
while winnr <= tabpagewinnr(tab, '$')
  " ...
  let winnr += 1
endwhile
```

There's still an edge case in that we could silently skip over a second status window if it moves out from under us, but I'm willing to live with that.

A true fix would utilize `get_winid()` to get a persistent reference to each window, but I'd rather wait till after dropping Vim 7 compatibility to roll that out.

ianhomer:
Yep - that's better.  I've updated branch and verified it fixes issue described in #1680

PR #1686, opened 2021-02-27
Fix bug where StageIgnore Gsplits an invalid fugitive-object

tim-clifford:
when using `1gI` or `1gi`, `s:StageIgnore` attempts to execute `Gsplit .gitignore`, which results in the error 
```
fatal: 0000000000000000000000000000000000000000 is neither a commit nor blob(master)
```
and subsequently
```
Error detected while processing function <SNR>56_StageIgnore:
line    9:
E21: Cannot make changes, 'modifiable' is off
```
using `split` instead of `Gsplit` fixes this, although I'm not sure if there is a more idiomatic fix?

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -3647,7 +3647,7 @@ function! s:StageIgnore(lnum1, lnum2, count) abort
     call extend(paths, info.relative)
   endfor
   call map(paths, '"/" . v:val')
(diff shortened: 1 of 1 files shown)


tpope:
This was fixed in https://github.com/tpope/vim-fugitive/commit/eb4d75efb621870393fe174e1647c585544eb732.

tim-clifford:
Sorry for wasting your time, I didn't update fugitive properly before opening this.

Issue #1702, opened 2021-03-17
Gedit: no fold found


mdedonno1337:
Hi,

When using the `:Gedit:` command and having the `set nofoldenable` line set in the vimrc file, when using the 'I' binding to go the the next hunk to stage, we have this error:

![Screenshot from 2021-03-17 14-49-51](https://user-images.githubusercontent.com/9901302/111479107-f5fb5000-8730-11eb-8df4-be8ffea7d360.png)

I've made a PR.

mdedonno1337:
Patch later on master, sorry about that ...

PR #1703, opened 2021-03-17
Enable local folding for the StageInline function

mdedonno1337:
This will patch the issue #1702.


mdedonno1337:
Patched after in mater, sorry about that.

Issue #1700, opened 2021-03-16
Display branch names in Gstatus


Gelio:
Hey,

While working with `:Gstatus` after a rebase I found it difficult to see which commits are on some existing branch (e.g. `origin/master`), and which are on my local branch. Example output:

![image](https://user-images.githubusercontent.com/889383/111308365-deea2e80-865a-11eb-96db-ebc6f02a86f3.png)

The view would certainly be more useful to me if there was a branch name displayed next to commits to which the branches point to, e.g.

![image](https://user-images.githubusercontent.com/889383/111308561-1eb11600-865b-11eb-99c4-82197bbd4ea4.png)

tpope:
First things first, I highly recommend using `origin/master` as your upstream, and setting `push.default` to current. This will give you an additional 2 sections based on your relation to `origin/master` which makes it a lot easier to grok what's going on.

That said, this is a good (if obvious) idea. The problem is that we're limited to plain text, so we'd need to make it something like `(origin/master)` in order to syntax highlight it, and even _that_ isn't good enough because it will match `(#1241)` in your screenshot. Best we can do is try to make it really weird to avoid false positives. There's a good chance this will happen some day but currently I have enough other experiments going on that I won't be adding this to the pile for the time being.

Gelio:
> First things first, I highly recommend using origin/master as your upstream, and setting push.default to current. This will give you an additional 2 sections based on your relation to origin/master which makes it a lot easier to grok what's going on.

That's an interesting approach. Yeah, this definitely makes it easier to see what's the status of the topic branch in relation to the upstream. Thanks for the tip!

---

Right, I see the potential problems. I guess using some obscure characters (I'm thinking `«origin/master»`) would it less prone to false positives.

Out of curiosity, if it happens that I have a free afternoon, are you open to contributions in this area?

Anyway, thanks for the answer

tpope:
The initial implementation is not the bottleneck, but if you wanna start prototyping and get a feel for what works and what doesn't, have at it. Shouldn't be much more to it than updating `s:QueryLog()`, `s:FormatLog()`, and the syntax highlighting.

Gelio:
After some experimentation, I produced https://github.com/tpope/vim-fugitive/pull/1713

PR #1713, opened 2021-03-24
Show refs in log in Gstatus

Gelio:
This is my attempt to close https://github.com/tpope/vim-fugitive/issues/1700. I have followed your advice and modified the functions you mentioned. Writing in VimScript was a new experience for me and I appreciate the fact that the change was relatively simple.

The PR adds displaying refs (the output of `%d` from `git log`'s `format=pretty`) next to commit summaries in `Gstatus`.

It uses basic syntax highlighting for all refs.

## Examples 

image
image

### DIFF
--- a/syntax/fugitive.vim
+++ b/syntax/fugitive.vim
@@ -26,7 +26,8 @@ syn match fugitiveSymbolicRef /\.\@!\%(\.\.\@!\|[^[:space:][:cntrl:]\:.]\)\+\.\@
 syn match fugitiveHash /^\x\{4,\}\S\@!/ contained containedin=@fugitiveSection nextgroup=fugitiveRefs skipwhite
 syn match fugitiveHash /\S\@<!\x\{4,\}\S\@!/ contained
 syn match fugitiveRefs '()' contained conceal
-syn match fugitiveRefs /([^)]\+)/hs=s+1,he=e-1 contained
+syn match fugitiveRefs /([^)]\+)/hs=s+1,he=e-1 contained contains=fugitiveRef
+syn match fugitiveRef /\<[^[:space:][:cntrl:]]\+\>/ contained
 
 syn region fugitiveHunk start=/^\%(@@\+ -\)\@=/ end=/^\%([A-Za-z?@]\|$\)\@=/ contains=@fugitiveDiff containedin=@fugitiveSection fold
 
@@ -53,7 +54,7 @@ hi def link fugitiveStagedModifier Typedef
 hi def link fugitiveInstruction Type
 hi def link fugitiveStop Function
 hi def link fugitiveHash Identifier
-hi def link fugitiveRefs Function
+hi def link fugitiveRef Function
 hi def link fugitiveSymbolicRef Function
 hi def link fugitiveCount Number
 


autoload/fugitive.vim
```diff
@@ -1701,14 +1701,14 @@ function! s:ReplaceCmd(cmd) abort
...
+  let lines = s:LinesError(['log', '-n', '256', '--pretty=format:%h%x09%s%x09%D', a:refspec, '--'])[0]
   call map(lines, 'split(v:val, "\t")')
-  call map(lines, '{"type": "Log", "commit": v:val[0], "subject": v:val[-1]}')
+  call map(lines, '{"type": "Log", "commit": v:val[0], "subject": v:val[1], "ref": get(v:val, 2, "")}')
   return lines
 endfunction
 
 function! s:FormatLog(dict) abort
-  return a:dict.commit . ' ' . a:dict.subject
+  return a:dict.commit . ' ' . a:dict.subject .. ((len(a:dict.ref) > 0) ? (' «' . a:dict.ref . '»') : '')
```
Gelio, line 1711: Using a single dot for concatenation resulted in an error here

autoload/fugitive.vim
```diff
@@ -1701,14 +1701,14 @@ function! s:ReplaceCmd(cmd) abort
 endfunction
 
 function! s:QueryLog(refspec) abort
-  let lines = s:LinesError(['log', '-n', '256', '--pretty=format:%h%x09%s', a:refspec, '--'])[0]
+  let lines = s:LinesError(['log', '-n', '256', '--pretty=format:%h%x09%s%x09%D', a:refspec, '--'])[0]
   call map(lines, 'split(v:val, "\t")')
-  call map(lines, '{"type": "Log", "commit": v:val[0], "subject": v:val[-1]}')
+  call map(lines, '{"type": "Log", "commit": v:val[0], "subject": v:val[1], "ref": get(v:val, 2, "")}')
```
Gelio, line 1706: That `-1` looked suspicious to me. I did not find a reason why it was done that way, though. Was it to guard against VimScript errors when the commit message is empty? If yes, then I reckon we should use `get` here for `v:val[1]`, shouldn't we?

tpope: The subject is the only thing we're parsing that can contain arbitrary characters, so it should always come last. There's no need to resort to weird Unicode characters if you do it that way. And `-1` is always the last field.

autoload/fugitive.vim
tpope: What error? Fugitive still supports versions of Vim that don't have `..`.

autoload/fugitive.vim
Gelio: Oh, right, sure. Does that mean that `v:val` may be of arbitrary length, since the subject might contain the `\t` that we are using as the delimiter?

I'm asking since I'm wondering how to pluck the ref since it only appears in some lines. If I swap the order of subject and refs, I would simply do `"ref": len(v:val) > 2 ? v:val[1] : ""`, but that may lead to false positives when the subject line contains multiple `\t` and no refs (that's assuming we are using the same `\t` delimiter in the format string)

autoload/fugitive.vim
Gelio: Good to know. Weird, though. No error now that I am trying it out again. I'll go back to using a single dot then

autoload/fugitive.vim
tpope: You are correct that it appears subjects with tabs aren't correctly supported. I think a robust fix would be `join(v:val[1 : -1], "\t")`, plus `split(v:val, "\t", 1)` to handle adjacent tabs correctly. Then change `1` to `2` to accommodate your new field. Expect a conflict when I fix this upstream.

Gelio: Sure thing, sounds good. I'll wait for the upstream fix so as not to implement the same logic twice

tpope:
I think we should probably match `git log --oneline`, with the decorations up front in parentheses.  I think starting a commit message with a parenthetical is suitably weird that we can ignore it. A problem with using `«` and `»` is that Vim still supports ASCII and simple 8 bit encodings, and well even default to it in some cases.

One thing I don't like is that `(HEAD -> master, origin/master, origin/HEAD)` is extremely common and distractingly wide.

autoload/fugitive.vim
```diff
@@ -1701,14 +1701,14 @@ function! s:ReplaceCmd(cmd) abort
 endfunction
 
 function! s:QueryLog(refspec) abort
-  let lines = s:LinesError(['log', '-n', '256', '--pretty=format:%h%x09%s', a:refspec, '--'])[0]
+  let lines = s:LinesError(['log', '-n', '256', '--pretty=format:%h%x09%s%x09%D', a:refspec, '--'])[0]
```
tpope, line 1704: Use `%d` here as it's older. Parentheses can be stripped, or used wholesale in the format change I proposed.

autoload/fugitive.vim
Gelio: Done, I've implemented it as you suggested

autoload/fugitive.vim
Gelio: Good point. I've used the parenthesized refs directly, without stripping them only to reapply them later

Gelio:
> I think we should probably match git log --oneline, with the decorations up front in parentheses.

Right, I've done that in the latest version. 

> One thing I don't like is that (HEAD -> master, origin/master, origin/HEAD) is extremely common and distractingly wide.

The only thing that comes to my mind is to only show 2-3 refs, and add `...` if there are more, so it would be `(HEAD -> master, origin/master, ...)`. Not sure if that's better, though

syntax/fugitive.vim
```diff
@@ -23,8 +23,9 @@ syn match fugitiveDone /^done\>/ contained containedin=@fugitiveSection nextgrou
 syn match fugitiveStop /^stop\>/ contained containedin=@fugitiveSection nextgroup=fugitiveHash skipwhite
 syn match fugitiveModifier /^[MADRCU?]\{1,2} / contained containedin=@fugitiveSection
 syn match fugitiveSymbolicRef /\.\@!\%(\.\.\@!\|[^[:space:][:cntrl:]\:.]\)\+\.\@<!/ contained
-syn match fugitiveHash /^\x\{4,\}\S\@!/ contained containedin=@fugitiveSection
+syn match fugitiveHash /^\x\{4,\}\S\@!/ contained containedin=@fugitiveSection nextgroup=fugitiveRefs
 syn match fugitiveHash /\S\@<!\x\{4,\}\S\@!/ contained
+syn match fugitiveRefs /([^)]\+)/hs=s+1,he=e-1 contained containedin=@fugitiveSection
```
Gelio, line 28: Unfortunately, this applies to all parentheses in the subject 😟 
image

I'm a vim-syntax-authoring noob, so hints are welcome

tpope: You'll want to drop the `containedin` and use `nextgroup` on the hash.

tpope:
Poking around in some histories I occasionally see messages starting with `(Hopefully) fix` or `(trivial)`. I don't have a solution, but I would be inclined to explore using syntax concealing to throw an extra character or two in there to disambiguate.

> > One thing I don't like is that (HEAD -> master, origin/master, origin/HEAD) is extremely common and distractingly wide.
> 
> The only thing that comes to my mind is to only show 2-3 refs, and add `...` if there are more, so it would be `(HEAD -> master, origin/master, ...)`. Not sure if that's better, though

I'm inclined to explore solutions of the form "if we're in the `Unpulled from origin/master` section, we don't need to show `origin/master` on the first commit". Also I think `HEAD` and `origin/HEAD` are pretty low value.

syntax/fugitive.vim
Gelio: After scratching my head for a while, `skipwhite` was also necessary.

I'll skip updating the branch with this intermediate step since I'll work on concealing and this will likely change anyway

tpope: Concealing has tradeoffs that may or may not be acceptable, so I wouldn't put much effort into that beyond proof of concept.

Gelio: I see. Ok then, I'll share the fixed version without concealing for now

Gelio:
I am experimenting with concealing. So far I have managed to add a `\t` at the beginning of the refs parentheses to detect false positives, and it looks to be working:

image

Disregard the fact that the parentheses themselves are not shown. I will add them back in the final version.

The thing I am worried about is that because I have added the concealed text, when using `h` or `l` to move horizontally in the line, those characters are still there and it takes an additional step to go over them, and they are not shown. Is this expected? Am I missing some core function? I've set `conceallevel=3` and `concealcursor=nc`

Also, if you've got a better idea for a character to disambiguate the refs parentheses (I've used `\t`), let me know

> I'm inclined to explore solutions of the form "if we're in the `Unpulled from origin/master` section, we don't need to show `origin/master` on the first commit". Also I think `HEAD` and `origin/HEAD` are pretty low value.

I suppose I can pass a list of refs to ignore when `s:QueryLog` is called

Gelio:
FYI, the latest solution has concealing and does not highlight subjects that begin with parentheses

image

tpope:
> The thing I am worried about is that because I have added the concealed text, when using `h` or `l` to move horizontally in the line, those characters are still there and it takes an additional step to go over them, and they are not shown. Is this expected? Am I missing some core function? I've set `conceallevel=3` and `concealcursor=nc`

That's the correct behavior, and it's the biggest of the "tradeoffs" I alluded to. Another is that if you copy the line and paste it to another buffer, you'll see the concealed characters.
 
> Also, if you've got a better idea for a character to disambiguate the refs parentheses (I've used `\t`), let me know

One idea is to prepend `"() "` to any commit message that starts with `(` and lacks decorations, and then conceal only empty parentheses. Then the cursor weirdness would be limited to weird commit messages instead of every decorated line.

> > I'm inclined to explore solutions of the form "if we're in the `Unpulled from origin/master` section, we don't need to show `origin/master` on the first commit". Also I think `HEAD` and `origin/HEAD` are pretty low value.
> 
> I suppose I can pass a list of refs to ignore when `s:QueryLog` is called

Ignoring the parameter `a:refspec` is probably a good start. Whatever crude hacks you come up with for this are fine. Let me reiterate that the bottleneck probably won't be implementation, it's the design. Aim to get it good enough for you to use, and use it a while, and report back with your thoughts.

And just to make sure expectations are clear: I won't be merging until after the next stable release at the very earliest, and there are no guarantees I won't bail entirely.

Gelio:
> One idea is to prepend `"() "` to any commit message that starts with `(` and lacks decorations, and then conceal only empty parentheses. Then the cursor weirdness would be limited to weird commit messages instead of every decorated line.

Good idea, I'll investigate it tomorrow



> Ignoring the parameter `a:refspec` is probably a good start. Whatever crude hacks you come up with for this are fine.

I've added a separate parameter to avoid hacks with splitting the `a:refspec` by `..`. I've also added hiding `HEAD`. `origin/HEAD` can be hidden if we have a list of all remotes, but I'm not sure that's worth the time investment at this phase



> Aim to get it good enough for you to use, and use it a while, and report back with your thoughts.
> 
> And just to make sure expectations are clear: I won't be merging until after the next stable release at the very earliest, and there are no guarantees I won't bail entirely.

Sure thing, thanks for all the guidance and open communication

tpope:
> I've added a separate parameter to avoid hacks with splitting the `a:refspec` by `..`. I've also added hiding `HEAD`. `origin/HEAD` can be hidden if we have a list of all remotes, but I'm not sure that's worth the time investment at this phase

Sounds like a job for `--decorate-refs-exclude=` or maybe `--decorate=full`, but you're probably right about the time investment.

Gelio:
> One idea is to prepend `"() "` to any commit message that starts with `(` and lacks decorations, and then conceal only empty parentheses. Then the cursor weirdness would be limited to weird commit messages instead of every decorated line.

That was a great tip, I've just implemented that:

image
image

I've pushed that as a separate commit. If you'd like to merge the PR any time in the future, let me know if you'd prefer them squashed.



> Sounds like a job for `--decorate-refs-exclude=` or maybe `--decorate=full`

Thanks for the pointers again, I'll see what I can do.

---

Overall, I'd like to thank you for the support so far. I feel I have learned a ton with this PR 🙂

Gelio:
> > Sounds like a job for `--decorate-refs-exclude=` or maybe `--decorate=full`
> 
> Thanks for the pointers again, I'll see what I can do.

After a few minutes of testing, looks like `--decorate-refs-exclude` is essentially what I have implemented as part of this PR, but it requires adding additional stars (or prefixing the remote refs with `refs/remotes`) to work:

```sh
git log --decorate --decorate-refs-exclude='*origin/master' --all
git log --decorate --decorate-refs-exclude='refs/remotes/origin/master' --all
```

I did not come up with a useful case for `--decorate=full`

---

I have also experimented a bit with syntax highlighting only the refs, and not the commas in the refs list, but failed again. It either did not highlight anything at all or highlighted the whole line. I guess I'll leave it be for now

tpope:
> After a few minutes of testing, looks like `--decorate-refs-exclude` is essentially what I have implemented as part of this PR, but it requires adding additional stars (or prefixing the remote refs with `refs/remotes`) to work:
> 
> ```shell
> git log --decorate --decorate-refs-exclude='*origin/master' --all
> git log --decorate --decorate-refs-exclude='refs/remotes/origin/master' --all
> ```

I think we wanna exclude `master` in the unpushed section and `origin/master` in the unpulled section, since those are _always_ gonna be the top line. That is, we always wanna exclude the first half of `a..b`. The caller of `s:QueryLog()` is in a position to pass in the `refs/..` part. `head` is always `refs/heads/`, while `pull` is usually `refs/remotes`, unless `fetch_remote` is `.`, in which case it's also `refs/heads/`, and same for `push`. The conditionals that define `fetch_remote` and `push_remote` are probably a good place to add `fetch_ref` and `push_ref`.

> I have also experimented a bit with syntax highlighting only the refs, and not the commas in the refs list, but failed again. It either did not highlight anything at all or highlighted the whole line. I guess I'll leave it be for now

The fix is another level of `contains`. Don't highlight `fugitiveRefs`, create a `fugitiveRef` that's contained in it and highlight that. This part isn't very important right now.

Gelio:
> I think we wanna exclude `master` in the unpushed section and `origin/master` in the unpulled section, since those are _always_ gonna be the top line. That is, we always wanna exclude the first half of `a..b`. The caller of `s:QueryLog()` is in a position to pass in the `refs/..` part. `head` is always `refs/heads/`, while `pull` is usually `refs/remotes`, unless `fetch_remote` is `.`, in which case it's also `refs/heads/`, and same for `push`. The conditionals that define `fetch_remote` and `push_remote` are probably a good place to add `fetch_ref` and `push_ref`.

Thanks for a detailed answer. The rules make sense, I'll try to incorporate it sometime soon

> The fix is another level of `contains`. Don't highlight `fugitiveRefs`, create a `fugitiveRef` that's contained in it and highlight that. This part isn't very important right now.

I was trying to implement exactly this earlier today, but I think I overcomplicated it. I've managed to do it starting fresh. Thanks!

image

Dezzj (approved):

Scutua (approved):

Issue #1644, opened 2020-12-09
Gblame - E484: Can't open file - Powershell Core/pwsh


Acelya-9028:
Hello,

I use Neovim with pwsh as shell.

```
set shell=pwsh shellpipe=\| shellxquote=
set shellcmdflag=-NoLogo\ -NoProfile\ -ExecutionPolicy\ RemoteSigned\ -Command
set shellredir=\|\ Out-File\ -Encoding\ UTF8
```
```
:set shell? shellcmdflag?
shell=pwsh
shellcmdflag=-NoLogo -NoProfile -ExecutionPolicy RemoteSigned -Command

:echo tempname()
C:\Users\K2-18b\AppData\Local\Temp\nvimXOWntw\7

:echo fnamemodify(tempname(), ':p:h')
C:\Users\K2-18b\AppData\Local\Temp\nvimXOWntw

:echo fnamemodify(fnamemodify(tempname(), ':h'), ':p')
C:\Users\K2-18b\AppData\Local\Temp\nvimXOWntw\

:!git blame init.vim
OK!

:echo system('git blame init.vim')
OK!

:Gblame
ligne 102:
E484: Can't open file...
```

![gblame_error](https://user-images.githubusercontent.com/45818681/101643682-aa038580-3a34-11eb-8bb1-1b9eb3b8fe16.png)

Gstatus works well.
Gblame works when I comment all shell related variables in my vim config.

I followed this [discussion](https://github.com/tpope/vim-fugitive/issues/435) and test `set noshelltemp` and Gblame works.

So Gblame works when it uses a pipe but not a file.
Can I get Gblame to work with a file instead a pipe or it doesn't matter?

Thank you for your help.

EDIT: I use W10
```
Nom du système d’exploitation:              Microsoft Windows 10 Entreprise LTSC
Version du système:                                10.0.17763 N/A version 17763
```

gibwar:
PowerShell Core (pwsh) encodes using UTF-8 by default, so the `Out-File` isn't necessary any more. I run with the following settings and `:Gblame` works fine (my profile skips a lot of stuff when running under Vim, so I don't include `-NoProfile` but it doesn't impact anything here that I tested, ie. it worked with and without `-NoProfile`).

```vim
set shell=pwsh
set shellcmdflag=-NonInteractive\ -ExecutionPolicy\ RemoteSigned\ -Command
set shellquote=\"
set shellxquote=
set shellpipe=>%s\ 2>&1
[...]

PR #1783, opened 2021-07-03
Make compatible with latest VIM PowerShell support.

mrdubya:
PowerShell shell support has been improved as of VIM 8.2.3079 for Windows, MacOS, and Linux. This means that some changes are now required to work correctly.
Note that git status -bz under PowerShell appends a CR-LF pair which caused the summary buffer to always think there was 1 unnamed unstaged file and 1 unnamed staged file.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -460,7 +460,7 @@ function! s:BuildEnvPrefix(env) abort
   let env = items(a:env)
   if empty(env)
     return ''
-  elseif &shellcmdflag =~# '-Command'
+  elseif &shell =~? '\%(powershell\|pwsh\)\%(\.exe\)\=$'
     return join(map(env, '"$Env:" . v:val[0] . " = ''" . substitute(v:val[1], "''", "''''", "g") . "''; "'), '')
   elseif s:winshell()
     return join(map(env, '"set " . substitute(join(v:val, "="), "[&|<>^]", "^^^&", "g") . "& "'), '')
@@ -1532,9 +1532,11 @@ function! s:TempCmd(out, cmd) abort
   try
     let cmd = (type(a:cmd) == type([]) ? fugitive#Prepare(a:cmd) : a:cmd)
     let redir = ' > ' . a:out
-    if (s:winshell() || &shellcmdflag ==# '-Command') && !has('nvim')
+    if s:winshell() && !has('nvim')
       let cmd_escape_char = &shellxquote == '(' ?  '^' : '^^^'
       return s:SystemError('cmd /c "' . s:gsub(cmd, '[<>%]', cmd_escape_char . '&') . redir . '"')
+    elseif &shell =~? '\%(powershell\|pwsh\)\%(\.exe\)\=$'
+      return s:SystemError(&shell . ' ' . &shellcmdflag . ' ' . s:shellesc(cmd . redir))
     elseif &shell =~# 'fish'
       return s:SystemError(' begin;' . cmd . redir . ';end ')
     else


autoload/fugitive.vim
```diff
@@ -460,7 +460,7 @@ function! s:BuildEnvPrefix(env) abort
   let env = items(a:env)
   if empty(env)
     return ''
-  elseif &shellcmdflag =~# '-Command'
+  elseif &shell =~# '\(powershell\|pwsh\)\(.exe\)\?$'
```
tpope, line 463: Let's change the pattern to `'\%(powershell\|pwsh\)\(\.exe\)\=$` to match my preferred styling and the operator to `=~?` because the `PowerShell` capitalization seems common.

autoload/fugitive.vim
```diff
@@ -1963,12 +1965,14 @@ function! fugitive#BufReadStatus() abort
             let files = file
           endif
           let sub = matchstr(line, '^[12u] .. \zs....')
-          if line[2] !=# '.'
-            call add(staged, {'type': 'File', 'status': line[2], 'filename': files, 'submodule': sub})
-          endif
-          if line[3] !=# '.'
-            let sub = matchstr(line, '^[12u] .. \zs....')
-            call add(unstaged, {'type': 'File', 'status': get({'C':'M','M':'?','U':'?'}, matchstr(sub, 'S\.*\zs[CMU]'), line[3]), 'filename': file, 'submodule': sub})
+	  if len(sub)
```
tpope, line 1968: Is that a hard tab?

tpope (changes requested):

tpope:
Thanks for this, and for the upstream Vim changes. After #1644 I peaked at the Vim source and was surprised to discover there wasn't even an _attempt_ to address PowerShell.

autoload/fugitive.vim
mrdubya: No problem. Just found an issue with pwsh on Windows so a little more work needed.

autoload/fugitive.vim
mrdubya: Shucks, yes. It always slips through the cracks jumping across projects as to what and how they are enforced (e.g. use of modelines).

mrdubya:
I think most Windows users think invoking VIM from PowerShell results in using PowerShell as the shell from VIM. This is wrong, VIM will default to cmd.exe on Windows. Then again, using PowerShell as the shell in VIM is not a great environment as filters do not work, PowerShell does not support input redirection, at least not as most people expect it to work in a shell. But we digress.

mrdubya:
Some fun in TempCmd() - needed to craft a shell invocation to grab the redirected output, working around any existing redirected output from shellredir as happens with pwsh on Windows. Can't do it in SystemError() as that is used from elsewhere that can cope with the normal redirected output.

autoload/fugitive.vim
tpope: Err, use `\%(` for the second `\(` too, I overlooked that.

autoload/fugitive.vim
```diff
@@ -1963,12 +1965,14 @@ function! fugitive#BufReadStatus() abort
             let files = file
           endif
           let sub = matchstr(line, '^[12u] .. \zs....')
-          if line[2] !=# '.'
-            call add(staged, {'type': 'File', 'status': line[2], 'filename': files, 'submodule': sub})
-          endif
-          if line[3] !=# '.'
-            let sub = matchstr(line, '^[12u] .. \zs....')
-            call add(unstaged, {'type': 'File', 'status': get({'C':'M','M':'?','U':'?'}, matchstr(sub, 'S\.*\zs[CMU]'), line[3]), 'filename': file, 'submodule': sub})
+          if len(sub)
```
tpope, line 1968: The fundamental problem is that `line` is empty? Let's change the `elseif line[0] !=# '#'` conditional to `elseif len(line) && line[0] !=# '#'` to more accurately convey what's going on (and greatly reduce the diff size to boot).

tpope:
My work-in-progress jobs branch already has `s:TempCmd()` re-implemented. So if that's the only sticking point, we can go ahead with merging this and I can escalate getting that piece of my branch merged in as well.

tpope:
Check the `job-temp-files` branch I just pushed if you want to verify that everything combined together works as expected.

autoload/fugitive.vim
mrdubya: The line has a single character in it, LF. Do you want to handle the specific case or add in remaining line length check as well?

mrdubya:
Quick comment on your wip branch, `s:PrepareJob()` defaults to cmd.exe for win32 platforms. This gets into the realm of do you honor the users shell setting, or use a shell that should work regardless?

autoload/fugitive.vim
tpope: That's a bit grosser, but `elseif line[0] =~# "[^#\n]"` or something similar seems like the way to go. Or clean it up at the source in `s:NullError()` if there's a natural way to do that. On that note, check that `:echo FugitiveConfig()` doesn't have any stray newlines, as that's the other big consumer of `s:NullError()`.

tpope: It looks like Git commands with a `-z` flag always terminate with a final null byte, so just categorically stripping off a final `\n` seems safe.

tpope: Scratch all of that, I fixed the hell out of this on master. You should be able to toss this hunk away and pull.

tpope:
That `cmd.exe` is only a fallback for older Git versions. And since implementing that I've made 1.8.5 mandatory so it can be ditched entirely.

autoload/fugitive.vim
mrdubya: How about the following?
```
-        elseif line[0] !=# '#'
+        elseif line[0] !~# '[#[:space:]]'
````

Comment (tpope) [reply autoload/fugitive.vim:1968]: Great but my previous comment renders it moot.

**Comment by mrdubya** [comment]
Ignore the merge please. I'll look into this some more the next few days.

**Comment by mrdubya** [comment]
I have kicked the tyres with a merge of my changes and your `job-temp-files` branch and didn't seen any obvious problems. I'll run with it for the next week to see if anything pops up.

File: autoload/fugitive.vim [hunk 4]
```diff
@@ -1949,7 +1951,7 @@ function! fugitive#BufReadStatus() abort
           let props[prop[1]] = prop[2]
         elseif line[0] ==# '?'
           call add(untracked, {'type': 'File', 'status': line[0], 'filename': line[2:-1]})
-        elseif line[0] !=# '#'
+        elseif line[0] !~# '[#[:space:]]'
```
tpope, line 1954: Doesn't ae6f84adf3240e007f2be077d103781f65d4a437 render this unnecessary?

mrdubya: It does, my fork was behind. An update and rebase removes it. With the reworking of `s:TempCmd()` the change comes down to just the one line change in `s:BuildEnvPrefix()`.

tpope: So if we postpone merging my `s:TempCmd()` reworking, how well does this patch work on its own? Is _your_ change to `s:TempCmd()` adequate, or are there still problems?

autoload/fugitive.vim
```diff
@@ -460,7 +460,7 @@ function! s:BuildEnvPrefix(env) abort
   let env = items(a:env)
   if empty(env)
     return ''
-  elseif &shellcmdflag =~# '-Command'
+  elseif &shell =~? '\%(powershell\|pwsh\)\(\.exe\)\=$'
```
tpope, line 463: Please make this '\%(powershell\|pwsh\)\%(\.exe\)\=$'. I always favor `\%(` when possible for ease of reasoning. (My initial requested change overlooked the second capture group.)

autoload/fugitive.vim
```diff
@@ -1532,9 +1532,11 @@ function! s:TempCmd(out, cmd) abort
   try
     let cmd = (type(a:cmd) == type([]) ? fugitive#Prepare(a:cmd) : a:cmd)
     let redir = ' > ' . a:out
-    if (s:winshell() || &shellcmdflag ==# '-Command') && !has('nvim')
+    if s:winshell() && !has('nvim')
       let cmd_escape_char = &shellxquote == '(' ?  '^' : '^^^'
       return s:SystemError('cmd /c "' . s:gsub(cmd, '[<>%]', cmd_escape_char . '&') . redir . '"')
+    elseif &shell =~? '\%(powershell\|pwsh\)\(\.exe\)\=$'
```
tpope, line 1538: This one too.

autoload/fugitive.vim
tpope: Let me ask a simpler question: Is there any reason not to go ahead and merge this? If it's a step in the right direction, I'd rather go ahead and get it in, even if there are caveats.

mrdubya: Apologies, it has been a busy week for me. Yes, without your `s:TempCmd()` reworking my patch was adequate.  Thanks for merging the change.

PR #1786, opened 2021-07-08
Fix ternary expression in s:NullError

### LINKED ISSUE #1785: Issue completion fails with includeIf directives in gitconfig
Issue and collaborator completion fails with `includeIf` directives in git's config. 

I discovered caching of entries is not supported for `includeIf` directives (d48fc4a5) but I suppose we want to fail silently if encountered ([return on exec_error](https://github.com/tpope/vim-fugitive/blob/b498607aa79cc7099d02c30ae72812c245cc7091/autoload/fugitive.vim#L655-L657))?

I believe this is due to a syntax error in `s:NullError` (an incomplete ternary expression): [s:NullError](https://github.com/tpope/vim-fugitive/blob/b498607aa79cc7099d02c30ae72812c245cc7091/autoload/fugitive.vim#L557)

eliasnorrby:
Only the leading part was removed in ae6f84a.

Fixes #1785

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -554,7 +554,7 @@ endfunction
 function! s:NullError(...) abort
   let [out, exec_error] = s:SystemError(call('fugitive#Prepare', a:000))
   if exec_error
-    return [[], substitute(out, "\n$", "", "") : '', exec_error]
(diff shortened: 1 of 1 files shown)


tpope:
Whoops!

PR #1803, opened 2021-07-30
Add :Gx, :Gwa, :Gwqa and :Gxa commands

Felixoid:
Hey dear Tim,
I'd like to improve UX a little bit.

In the patch, I've added `:Gwa` command that I miss so much in the daily work. At the same time, to save some keystrokes and provide the same behavior as vim, I've added `:Gx[a]` commands.

Hopefully, you find it useful!

### DIFF
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -166,6 +166,11 @@ that are part of Git repositories).
 :Gwq! [path]            Like |:Gwrite|! followed by |:quit|! if the write
 :Gx! [path]             succeeded.
 
+                                                *:Gwa* *:Gwqa* *:Gxa*
+:Gwa[!]                 The commands work accordingly *:wa* *:wqa* and *:xa*
+:Gwqa[!]                commands. Each stages the changes in all active
+:Gxa[!]                 buffers. *:Gwqa* and *:Gxa* additionally |:quit|.
+
                                                 *:Gdiffsplit*
 :Gdiffsplit [object]    Perform a |vimdiff| against the given file, or if a
                         commit is given, the current file in that commit.
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -147,24 +147,24 @@ that are part of Git repositories).
 :{range}Gread! [args]   |:read| the output of a Git command after {range}.
 
                                                 *:Gwrite* *fugitive-:Gw*
-:Gwrite                 Write to the current file's path and stage the results.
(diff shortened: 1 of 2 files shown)


tpope:
What happens when you call these while some non-Fugitive buffers are open?

Felixoid:
Thank you for asking. I've checked and it doesn't look to work as expected. Looks like the first `exe "wa"` is ignored. Let me work it around a little bit.

Felixoid:
So, here's a terminal record for the testing session, it can be replayed with `asciinema play gwa.txt` 
[gwa.txt](https://github.com/tpope/vim-fugitive/files/6911886/gwa.txt)

The only one side effect I could find is replacing some spaces in the help page buffer with tabs.

tpope:
Your asciinema also shows several `fugitive: file does not belong to a Git repository` errors too, presumably due to the help buffer. And help buffers are just one of numerous possible problematic buffers. The quickfix list, terminal buffers, readonly files, directory buffers (netrw), `http://` buffers, and files in `.gitignore` are the first handful that come to mind. I would expect a robust implementation to be significantly larger than a one liner.

But even with a robust implementation, I'm not sure this is a good idea. If I have files open from multiple repositories, should `:Gwa` stage all of them? Staging files across a bunch of unrelated repositories strikes me as more dangerous than helpful. Should we stage untracked files? If so, there's `:Git add -A`, and if not, there's `:Git add -u`. I don't think `:Gwa` adds much here.

Felixoid:
Thank you for feedback. It makes sense, especially with `:G add -u`

Would you consider meeting only first commit, maybe?

tpope:
From the help:

```
:[range]x[it][!] [++opt] [file]
			Like ":wq", but write only when changes have been
			made.
```

If no changes have been made, should we still stage the file? The fact there is no obvious answer to this question inclines me not to add the command.

tpope:
Now that I think about it, your patch doesn't even implement the normal distinction between `:wq` and `:x`. I don't think Fugitive should provide a `:Gx` that is unrelated to `:x`, especially since the only benefit seems to be one less character to type.

Felixoid:
> Now that I think about it, your patch doesn't even implement the normal distinction between `:wq` and `:x`. I don't think Fugitive should provide a `:Gx` that is unrelated to `:x`, especially since the only benefit seems to be one less character to type.

Quite honest. Thank you for your feedback!

PR #1814, opened 2021-08-19
Remove unnecessary `echo`

bobwhitelock:
This `echo`, added in https://github.com/tpope/vim-fugitive/commit/286bf9096de0d74a5b663f2c6d4ae879ef97d93b, changed the behaviour of `Ggrep` for me. It means every line found is echoed, and I need to scroll to the bottom of this output (when there are a lot of results) in order to be able to hit enter and have my quickfix window populated with the grep results, whereas previously I could just hit enter to have this happen immediately.

This commit removes this to restore the previous behaviour; I'm guessing this was for debugging but was accidentally left here - unless I'm missing some reason this is actually needed.

Thanks!

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -5136,7 +5136,6 @@ function! fugitive#LogComplete(A, L, P) abort
 endfunction
 
 function! s:GrepParseLine(options, dir, line) abort
-  echo a:line
   let entry = {'valid': 1}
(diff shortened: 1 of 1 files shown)


tpope:
It was not for debugging; it was to match `:grep` which also dumps all the output for you to see. But now that you mention it, `:grep` autoscrolls to the bottom while the new `:Ggrep` doesn't. Dang, not sure what to do about this.

bobwhitelock:
Thanks for fixing!

PR #1837, opened 2021-09-13
Use --filters in cat-file whenever possible

### LINKED ISSUE #1509: git-crypt support?
If I try to use `:Gdiffsplit` in a project that is encrypted with git-crypt, it always tries to diff with an encrypted file.

If i do a git diff or a git difftool on the commandline it works properly.
Is it possible to configure fugitive to work properly with git crypt?

YorikSar:
Starting with Git 2.11 cat-file provides --filters option that can be
specified instead of "blob" type. This option tells Git to use all
filters specified in .gitattributes.

This enables diffs for files with filters, for example, git-crypt.

Closes #1509

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -2109,7 +2109,8 @@ function! s:BlobTemp(url) abort
   endif
   if commit =~# '^\d$' || !filereadable(tempfile)
     let rev = s:DirRev(a:url)[1]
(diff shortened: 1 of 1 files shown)


justinmk:
> This option tells Git to use all filters specified in .gitattributes.

Is this a security risk, i.e. should it be an explicit opt-in per repository? I suppose some `:Git` commands could be considered "opt in" by virtue of their usage, but not others which are normally read-only.

tpope:
Correct, not a security risk.

PR #1843, opened 2021-09-23
Fix missing args variable in RemoteUrl

Frederick888:
It seems to be an issue introduced by 1352646.


tpope:
Thanks for the heads up.

PR #1898, opened 2021-12-01
Fix detection of worktrees specified in config.worktree (extensions.worktreeConfig)

liskin:
This adds support for bare repos with extensions.worktreeConfig=true and core.worktree configured for those worktrees. Specifically, the following two setups are now supported:

1. /path/to/repo.git/config: extensions.worktreeConfig=true
   /path/to/repo.git/config.worktree: core.worktree=/main/worktree/location

2. /path/to/repo.git/config: extensions.worktreeConfig=true
   /path/to/repo.git/worktrees/wt1/commondir: ../..
   /path/to/repo.git/worktrees/wt1/gitdir: .
   /path/to/repo.git/worktrees/wt1/config.worktree: core.worktree=/other/worktree/location

I've used this setup with my dotfiles repo for a while, with the main worktree being /home/tomi, plus an additional / worktree for the root branch. Fugitive now works in this setup (I only tested :G without args).

### DIFF
--- a/plugin/fugitive.vim
+++ b/plugin/fugitive.vim
@@ -320,9 +320,18 @@ function! s:Tree(path) abort
   endif
   if !has_key(s:worktree_for_dir, dir)
     let s:worktree_for_dir[dir] = ''
-    let config_file = dir . '/config'
+    let commondir = fugitive#CommonDir(dir)
+    if empty(commondir)
+      let commondir = dir
+    endif
+    let config_file = commondir . '/config'
+    let worktree_config_file = dir . '/config.worktree'
     if filereadable(config_file)
       let config = readfile(config_file,'',10)
+      let ext_wtc_config = filter(config,'v:val =~# "^\\s*worktreeConfig *= *\\%(true\\|yes\\|on\\|1\\)$"')
+      if len(ext_wtc_config) == 1 && filereadable(worktree_config_file)
+        call extend(config, readfile(worktree_config_file,'',10))
+      endif
       let wt_config = filter(copy(config),'v:val =~# "^\\s*worktree *="')
       if len(wt_config) == 1
         let worktree = FugitiveVimPath(matchstr(wt_config[0], '= *\zs.*'))


plugin/fugitive.vim
```diff
@@ -320,9 +320,18 @@ function! s:Tree(path) abort
...
-    let config_file = dir . '/config'
+    let commondir = fugitive#CommonDir(dir)
+    if empty(commondir)
+      let commondir = dir
+    endif
+    let config_file = commondir . '/config'
+    let worktree_config_file = dir . '/config.worktree'
     if filereadable(config_file)
       let config = readfile(config_file,'',10)
+      let ext_wtc_config = filter(config,'v:val =~# "^\\s*worktreeConfig *= *\\%(true\\|yes\\|on\\|1\\)$"')
```
tpope, line 331: `filter()` mutates, so omitting `copy()` here breaks all other uses of `core.worktree`.

plugin/fugitive.vim
```diff
@@ -320,9 +320,18 @@ function! s:Tree(path) abort
   endif
   if !has_key(s:worktree_for_dir, dir)
     let s:worktree_for_dir[dir] = ''
-    let config_file = dir . '/config'
+    let commondir = fugitive#CommonDir(dir)
+    if empty(commondir)
```
tpope, line 324: When will this be true?

plugin/fugitive.vim
liskin: Oh, I was wondering why it's there… :-)
Now I know!

plugin/fugitive.vim
liskin: Good question. `fugitive#CommonDir` returns empty string when the `a:dir . '/HEAD'` file is too small, and I don't really know when that is. :-/

liskin: I just though… well, it might happen, let's have a fallback.

tpope: This size check is just enough for `ref: refs/x`; it's the same sanity check Git uses internally.

I guess this could happen if the user calls `FugitiveWorkTree('/invalid/path')`. In that event, we should probably just bail with `return ''`.

liskin: `return s:worktree_for_dir[dir]` seemed more consistent with other code around, but I'll be happy to change it to `return ''`

tpope: Ideally we would cache `''` but it's not really an approved usage so probably not worth bothering with.

liskin: Is it not cached with the code as it is?

tpope: I missed the first line of the conditional; it is indeed cached.

tpope:
Even with your fixes, I believe the logic here is flawed because it erroneously tries to use the `core.worktree` from inside the common dir. I think we should only check for the common dir if neither `config` nor `gitdir` is found. A structure like this might help minimize the repetition:

```vim
let has_config_file = filereadable(config_file)
if !has_config_file && filereadable(dir . '/gitdir')
  " ...
else
  if has_config_file
    " ...
  else
    " ...
  endif
  " ...
endif
```

A priority here is minimizing IO, in particular avoiding other checks when `config` is present.

liskin:
Oh, you're right. I'll need to revisit this with a clear head, hopefully tomorrow. Thanks for the feedback so far!

liskin:
Didn't really get to it "tomorrow" as I had some trouble with my wrist and had to stay away from the computer for a while, but I just pushed another attempt which I think addresses your concerns.

(Rebased on top of master, the recent changes make my code a bit simpler, too! :-))

tpope:
The recent changes were extracted from my attempt actually. Try the `extensions-worktree-config` branch.

plugin/fugitive.vim
```diff
@@ -329,20 +329,34 @@ function! s:Tree(path) abort
...
-    else
+    let config_is_common = 0
+    if empty(config) && filereadable(dir . '/gitdir')
       let worktree = fnamemodify(FugitiveVimPath(get(s:ReadFile(dir . '/gitdir', 1), '0', '')), ':h')
       if worktree ==# '.'
         unlet! worktree
+        let commondir = fugitive#CommonDir(dir)
+        if !empty(commondir)
+          let config = s:ReadFile(commondir . '/config', 10)
+          let config_is_common = 1
```
tpope, line 340: `fugitive#CommonDir()` doesn't return an empty string for a simple repository, but rather the Git dir. Just like `git rev-parse --git-common-dir`.

liskin:
> The recent changes were extracted from my attempt actually. Try the `extensions-worktree-config` branch.

Oh, nice! I just tested this with the repos in question and it appears to work well.

liskin:
Thanks ♥

PR #1936, opened 2022-01-24
Add basic bisect subcommand completion

rbong:
This adds some very basic support for bisect completion.

* Completes bisect subcommand (ex. "good", "bad", "start", "reset", etc.) if it is not already specified
* Otherwise, completes revision

Bisect subcommands are hardcoded.
There's no way as far as I can see to automatically get bisect subcommands, even `git-completion.sh` manually specifies them.

The pattern to find the subcommand is not very robust, but it is easy to read so let me know if you require any changes or if it is good enough to start.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -3948,7 +3948,7 @@ function! fugitive#Complete(lead, ...) abort
     let results = ['--literal-pathspecs', '--no-literal-pathspecs', '--glob-pathspecs', '--noglob-pathspecs', '--icase-pathspecs', '--no-optional-locks']
   elseif empty(subcmd)
     let results = s:CompletableSubcommands(dir)
-  elseif a:0 ==# 2 && subcmd =~# '^\%(commit\|revert\|push\|fetch\|pull\|merge\|rebase\)$'
+  elseif a:0 ==# 2 && subcmd =~# '^\%(commit\|revert\|push\|fetch\|pull\|merge\|rebase\|bisect\)$'
     let cmdline = substitute(a:1, '\u\w*\([! ] *\)' . subcmd, 'G' . subcmd, '')
     let caps_subcmd = substitute(subcmd, '\%(^\|-\)\l', '\u&', 'g')
     return fugitive#{caps_subcmd}Complete(a:lead, cmdline, a:2 + len(cmdline) - len(a:1), dir, root)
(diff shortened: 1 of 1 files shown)


tpope:
I think `fugitive#FetchComplete()` and `s:CompleteRemote()` should be the model here, as this is another use case that special cases the first argument. The provided implementation looks functionally sound but invents new idioms when old ones will do.

rbong:
I have updated this PR to use the pattern from `s:CompleteRemote()`.

I am not sure if you'd like me to follow `fugitive#FetchComplete()` and `s:CompleteRemote()` more closely. As far as consistency goes, right now it follows a similar idea to `fugitive#CommitComplete()`, where there are subcommand-specific completions handled before falling back on `s:CompleteSub()` and `s:CompleteRevision()`. If this is not what you were thinking, let me know.

tpope:
> I have updated this PR to use the pattern from `s:CompleteRemote()`.
> 
> I am not sure if you'd like me to follow `fugitive#FetchComplete()` and `s:CompleteRemote()` more closely.

Ideally yes, unless there's a catch I'm not seeing.

> As far as consistency goes, right now it follows a similar idea to `fugitive#CommitComplete()`, where there are subcommand-specific completions handled before falling back on `s:CompleteSub()` and `s:CompleteRevision()`. If this is not what you were thinking, let me know.

This special case is unfortunately necessary because the `--squash=` handling needs to be higher precedence than the generic `--git-completion-helper` handler. We don't need that for `bisect`.

We should also use `fugitive#CompleteObject()` rather than `s:CompleteRevision`. as some commands like `run` do accept file paths.

autoload/fugitive.vim
```diff
@@ -5247,6 +5247,17 @@ function! s:RebaseSubcommand(line1, line2, range, bang, mods, options) abort
   return {}
 endfunction
 
+" Section: :Git bisect
+
+function fugitive#BisectComplete(A, L, P, ...) abort
+  let bisect_subcmd = matchstr(a:L, '\u\w*[! ] *.\{-\}\s\@<=\zs[^-[:space:]]\S*\ze ')
+  if empty(bisect_subcmd)
+    let results = ['start', 'bad', 'new', 'good', 'old', 'terms', 'skip', 'next', 'reset', 'replay', 'log', 'run']
+    return filter(results, 'strpart(v:val, 0, strlen(a:A)) ==# a:A')
```
tpope, line 5256: Let's use `s:FilterEscape()` for this.

rbong:
I updated the branch again, I believe this mirrors the other completion functions a lot closer now.

PR #1980, opened 2022-04-21
fix documentation for blame map ~

AlterDepp:
In blame window `[count]~` means the [count]th first parent, not grandparent

### DIFF
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -70,8 +70,8 @@ that are part of Git repositories).
                         O     jump to patch or blob in new tab
                         p     jump to patch or blob in preview window
                         -     reblame at commit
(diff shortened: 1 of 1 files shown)


tpope:
Wow how on Earth did that get messed up? Let's ditch the "first" too, "4th first parent" makes no sense.

tpope:
Oh that plays weird against the next line. I'll think on it and figure something out.

AlterDepp:
The git documentation is weird, too. Maybe shorter is better.

tpope:
Yeah simplest is best. "Like HEAD" is misleading too, because it's "like" HEAD, but it's not actually HEAD, it's the commit you're already looking at.

PR #1991, opened 2022-05-13
Fixing Fugitive using Fugitive#Head instead FugitiveHead.

fgr1986:
Minimum changes to remove dependence on Fugitive#Head calls.


tpope:
?????

fgr1986:
@tpope Apologies, I mixed it up with a call from lightline.

PR #2013, opened 2022-06-12
Add: recreate commit message map

### LINKED ISSUE #1608: Feature Request: Restore last commit message
From time to time, `git commit` fails when called from fugitive, because of external reasons (e.g. I fail to properly sign the commit).
In that case I need to tell fugitive to call the same commit command again.
Unfortunately when I type `cc`, I will also have to retype the whole commit message again.
This can be a slight or huge pain.

So I think it would be good, if there is some way to restore the commit message of the previous failed commit call.
It seems to me, like it is already saved in `.git/COMMIT_EDITMSG`. So what about the following workflow:

- Type `cc`
- If `.git/COMMIT_EDITMSG` exists, reuse the non-commented part of it.
- `:wq`
- Call `git commit`.
    - If it succeeds, delete `.git/COMMIT_EDITMSG`
    - If it fails, keep `.git/COMMIT_EDITMSG`

ironhouzi:
`cd` for recreating the commit using the commit message from
`.git/COMMIT_EDITMSG` .

Fixes #1608

Hope this isn't too pushy. I just thought I'd try to create a fix, instead of submitting a feature request.

If there's a better way to add my own fugitive maps, then by all means please reject this PR if you don't see any value to it. But please let me know how to create my own fugitive maps, as I've tried to search the github issues and the help files for this information.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -7573,6 +7573,7 @@ function! fugitive#MapJumps(...) abort
     call s:Map('n', 'ca', ':<C-U>Git commit --amend<CR>', '<silent>')
     call s:Map('n', 'cc', ':<C-U>Git commit<CR>', '<silent>')
     call s:Map('n', 'ce', ':<C-U>Git commit --amend --no-edit<CR>', '<silent>')
+    call s:Map('n', 'cd', ':<C-U>Git commit --edit --file .git/COMMIT_EDITMSG<CR>', '<silent>')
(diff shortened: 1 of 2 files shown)


tpope:
I am going to pass for now. This won't work in work trees or submodules, where `.git` is a file rather than a directory. That could be fixed, but I'm going to take the fact Git doesn't provide a good way to do this as a sign that it's not really a recommended usage.

PR #2019, opened 2022-06-25
Don't silence cmap

tomtomjhj:


tpope:
Why tho?

tomtomjhj:
I'm not sure if it is intended, but the silenced `<C-R><C-F>` map doesn't show the value of `fugitive#Cfile()` in cmdline.
For example, this doesn't show anything in the cmdline and doesn't even move the cursor:
```
:G log<CR>:<C-R><C-F>
```
image

If I type `<BS>`, then it reveals the value.
image

tomtomjhj:
I'm not sure if it is intended, but the silenced `<C-R><C-F>` map doesn't show the value of `fugitive#Cfile()` in cmdline.
For example, this doesn't show anything in the cmdline and doesn't even move the cursor (tested with `vim --clean -c 'set rtp^=~/.vim/plugged/vim-fugitive' -c 'runtime plugin/fugitive.vim'`:
```
:G log<CR>:<C-R><C-F>
```
image

If I type `<BS>`, then it reveals the value.
image

PR #2025, opened 2022-07-21
Improve backward navigation maps (`[c`, `[/`, `[m`, `(`, `[[`)

yut23:
This PR fixes two bugs when navigating to previous items in fugitive buffers:

* in neovim (or vim with `set nostartofline`), `[/` and `[m` do nothing if the cursor isn't at the start of the line, while the others just move to the start of the current line
* `[c` doesn't expand the previous inline diff when the cursor is on the first hunk of a file

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -4609,6 +4609,7 @@ function! s:NextHunk(count) abort
 endfunction
 
 function! s:PreviousHunk(count) abort
+  normal! 0
   for i in range(a:count)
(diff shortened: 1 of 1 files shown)


tpope:
The `z` flag needs Vim 8.0. Can we get away with `normal! 0` beforehand instead?

yut23:
Yeah, I think that should work. I'll test it out.

tpope:
Thanks!

PR #2042, opened 2022-08-26
Replace vimcasts.org links with archive.org links

louwers:
Vimcasts.org is down, so I replaced the links in the README.


tpope:
Let's give it a week to make sure it isn't a temporary outage.

PR #2045, opened 2022-08-31
fix ingore start with ^ string

hanh090:
Gitlab commit can start commit string without `^` character, it leads to GBrowse in visual do not works. Change it to `.` for more flexible

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -7463,7 +7463,7 @@ function! fugitive#BrowseCommand(line1, count, range, bang, mods, arg, ...) abor
           endif
           call delete(blame_list)
           if !exec_error
-            let blame_regex = '^\^\x\+\s\+\zs\d\+\ze\s'
+            let blame_regex = '^.\x\+\s\+\zs\d\+\ze\s'
             if get(blame, 0) =~# blame_regex && get(blame, -1) =~# blame_regex
               let line1 = +matchstr(blame[0], blame_regex)
               let line2 = +matchstr(blame[-1], blame_regex)


tpope:
Look at the `else` clause. The only time there won't be a `^` is when there are unpushed changes.

hanh090:
Hi @tpope,

I was double check with an open Gitlab project and see that problem happen. Those are steps:
- Clone this project: https://gitlab.com/dennis/vim-easytags
- Open gitblame at file `README.md` by fugitive

Here is result
Screen Shot 2022-09-01 at 07 31 28

You can see the commit list was mix between `^` and without `^`

Therefore, problem is very possible?

tpope:
It works like this:

1. [Find the upstream ref for the remote we're blaming at.](https://github.com/tpope/vim-fugitive/blob/b411b753f805b969cca856e2ae51fdbab49880df/autoload/fugitive.vim#L7395-L7434).
2. [Find the `merge-base` commit of that ref and the current commit (usually HEAD).](https://github.com/tpope/vim-fugitive/blob/b411b753f805b969cca856e2ae51fdbab49880df/autoload/fugitive.vim#L7450-L7451)
3. [Write that commit to a file.](https://github.com/tpope/vim-fugitive/blob/b411b753f805b969cca856e2ae51fdbab49880df/autoload/fugitive.vim#L7454)
4. [Call `git blame -S` with that file.](https://github.com/tpope/vim-fugitive/blob/b411b753f805b969cca856e2ae51fdbab49880df/autoload/fugitive.vim#L7455-L7463)

If the lines were last changed _before_ the merge commit, the result will look like `^commit`. If not, it will be the unpushed commit that changed the line, which means the code selected has unpushed changes, and we can't reliably link to it.

If you're _sure_ you don't have anything unpushed, then either you have something weird in your local repository telling step 1 to look in the wrong place, or (unlikely but possible) step 1 has a bug in it.

PR #2069, opened 2022-10-20
Fix Git blame not being able to show commits on Windows

filcab:
If the username is long enough that the home directory gets truncated to an 8.3 path (e.g: C:/Users/ABCDEF~1), Git blame will fail on the silent :read command due to the ~ character.

Expand the path to a full path after the file exists, on Windows, to fix it.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -2550,6 +2550,13 @@ augroup END
 function! s:ReplaceCmd(cmd) abort
   let temp = tempname()
   let [err, exec_error] = s:StdoutToFile(temp, a:cmd)
+  if has('win32')
+    " on win32, we might get a short 8.3 path, which will have a ~ which will
+    " blow up on the read command further down. The following call to
+    " fnamemodify will expand the 8.3 paths *if* the file exists, so we only
+    " call it after writing the file in the line above
+    let temp = fnamemodify(temp, ':p')
+  endif
   if exec_error
     throw 'fugitive: ' . (len(err) ? substitute(err, "\n$", '', '') : 'unknown error running ' . string(a:cmd))
   endif


tpope:
> If the username is long enough that the home directory gets truncated to an 8.3 path (e.g: C:/Users/ABCDEF~1), Git blame will fail on the silent :read command due to the ~ character.

What error message does this give, exactly?

filcab:
```
Error detected while processing function <SNR>112_BlameCommit[21]..BufReadCmd Autocommands for "fugitive://*"..function fugitive#BufReadCmd[128]..function <SNR>112_BlameCommit[21]..BufReadCmd Autocommands for "fugitive://*"..function fugitive#BufReadCmd[80]..<SNR>112_ReplaceCmd:
line    8:
E480: No match: C:\Users\ABCDEF~1\AppData\Local\Temp\V3F97FE.tmp
Error detected while processing function <SNR>112_BlameCommit[21]..WinEnter Autocommands for "*"..function <SNR>110_OnBufferEnter:
line   10:
E117: Unknown function: UltiSnips#SnippetsInCurrentScope
```

If you have access to a Windows machine, I expect this to be reproable if you point `$TEMP` to a dir with a component longer that 8 chars.

filcab:
Ah. I added an `echom` just before the offending line, so the "line 8" should say "line 7" (it's the `silent exe ... 0read ++edit` line). Sorry about that.

tpope:
Check `:set wildignore?`, you have something in there blocking the file from being read.

filcab:
Interesting, indeed it was... `*~` in there. Unsure how I could have found this out from the `:read` and similar documentation, but I can't say I'm a vim expert.

Thank you very much!

PR #2077, opened 2022-11-08
Add keymap for commit without running git hooks

thepauljones:
`cn<Space>` will run `:Git commit --no-verify<CR>` 

I find myself typing this more than I want to and thought other might have the same.

I _love_ this tool, @tpope - thanks for creating it.


tpope:
Try `c<Space>-n` to achieve the same effect. The approach of "one map for every possible flag combination" has been pushed to it's breaking point. One day I hope to adopt a more flexible solution, but until then I'm not gonna dig the hole any deeper.

thepauljones:
Fair enough @tpope and thanks for the tip!

PR #2079, opened 2022-11-08
Allow X on section headers

### LINKED ISSUE #2063: Binding to throw away all (Unstaged / Untracked) changes
When I want to throw away all changes, I intuitively moved my cursor to the header 'Unstaged' and pressed 'X' but was disappointed that it did not work. Instead, we currently need to enter visual mode, `]` to the end of the paragraph and then 'X' - but it will take a long time (for example 20 seconds for about 300 files in my case) because it is saving each file.

It would be nice to have a quick way to do this.

TamaMcGlinn:
This reversibly throws away the changes.

Fixes #2063


Dezzj (commented):
Veiwed

PR #2086, opened 2022-11-23
Tighten match for commit under cursor

### LINKED ISSUE #2085: Checkout with coo fails for branch names that start with 4 or more numbers
Similar to the issue in #1756, using `coo` to open a branch under the cursor that starts with a number (e.g., `1111-feat`) fails with the error `fatal: invalid reference: 1111`. Looking at [this commit](https://github.com/tpope/vim-fugitive/commit/6ae064c5aa1cc24c78abc8d12ec236e99dc01316), we might consider adding a negative look-ahead.

I've minimally tested a change to `'\S\@<!\x\{4,\}\>\S\@!'`, but in the spirit of [don't be shy about asking before patching](https://github.com/tpope/vim-fugitive/blob/master/CONTRIBUTING.markdown), what do you think about this approach? I can test further. If it makes sense happy to open a PR (...or feel free to).

emilford:
Using `coo` to open a branch under the cursor that started with 4+ hex
digits (e.g., 1111-feat) would fail with the error `fatal: invalid reference: 
1111.`

This happened because the prior regex would consider these branch names
 as commits when determining whether to treat what was under the cursor
 as a branch or commit.

This changes the regex to use a negative look-ahead to further tighten
 the match for a commit and require that a match on 4+ hex digits also
not end with a non-hex digit.

Resolves #2085.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -7553,7 +7553,7 @@ function! s:SquashArgument(...) abort
   if &filetype == 'fugitive'
     let commit = matchstr(getline('.'), '^\%(\%(\x\x\x\)\@!\l\+\s\+\)\=\zs[0-9a-f]\{4,\}\ze \|^' . s:ref_header . ': \zs\S\+')
   elseif has_key(s:temp_files, s:cpath(expand('%:p')))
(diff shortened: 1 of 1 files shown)


tpope:
Thanks!

Issue #2070, opened 2022-10-31
Git blame and gdiffsplit do not work when a file has been moved


zdc13:
Git blame cannot browse to the file when it has been moved and modified in the same commit

Gdiffsplit does not work when the file has been moved between the diffed commits

To reproduce:
```
git init
echo test > myfile
git add . && git commit -m "First commit"
echo test >> myfile
mv myfile myfile_moved
git add .  && git commit -m "Second commit"
```
Open the `myfile_moved` in vim and type `:Gitdiffsplit HEAD^` : `fatal: path 'myfile_moved' exists on disk, but not in ...`

Open the file in vim and type `:Git blame`, go to the second line's commit and press P : `fatal: no such path myfile_moved in ...`

tpope:
> Open the `myfile_moved` in vim and type `:Gitdiffsplit HEAD^` : `fatal: path 'myfile_moved' exists on disk, but not in ...`

Fugitive translates `:Gdiffsplit HEAD^` to `:Gdiffsplit HEAD^:%`. I think this is a reasonable behavior. I don't think it's the _only_ reasonable behavior; certainly rename following can be helpful in some cases. But what _exactly_ do you propose? What Git command do we run that will spit out `myfile` given the appropriate input?

> Open the file in vim and type `:Git blame`, go to the second line's commit and press P : `fatal: no such path myfile_moved in ...`

The behavior here is less reasonable; the blame output is aware of the rename so we probably should be too. The `previous` line in `git blame --incremental` could probably serve as the implementation. It won't work with with a count (`2P`) but I am fine to limit this behavior to the count free version.

But I also consider `P` to be extremely low value compared to `<CR>`, which shows you the change in context. So low value, in fact, that I am taking this as an opportunity to remove it from the documentation. Thus I currently have no plans to pursue this improvement. A motivated contributor might start by looking at the `s:BlameJump()` function.

zdc13:
I use P to go back in history to know how a change developed. It allows me to use vim highlight, lsp and linters on the previous file, and to figure out where exactly a bug appeared. To me this as a lot more value than `<CR>`, which seems to only give the diff (just like a git show <commit> would do), which imo is useless to have inside vim itself. Am I misunderstanding the feature? P allows to go back more than a single blame and *then* to use `p` on the commit that caused an issue.

As for the git diff implementation, a more reasonable implementation might be to use a function to find the latest filename for a commit, like so:
```
[...]

PR #2115, opened 2023-02-19
Add "P" command to Git blame documentation

gregorias:
I found out about the command from https://advancedweb.hu/dive-into-git-history-with-fugitive-vim/, and I wish it was in the manual.


tpope:
This was deliberately removed from the documentation not that long ago as I felt it actively encouraged sub-optimal usage of the blame feature. See 09908d82efc602fde72ef02ae05a75953376f842 and #2070.

PR #2149, opened 2023-04-09
add optional formating to fugitive#Statusline()

greeschenko:
with g:statusline_fmt you can change git branch output format in status line 
 
 exemple:
 
 g:statusline_fmt = '%s'    return pure git branch name 
 
 or
 
  g:statusline_fmt = '>> %s'    return git branch name with custom symbols


tpope:
If you want a custom format you can write your own function.

Issue #2154, opened 2023-04-21
Feature: Display commit stats in order see what files are in a commit


kawas44:
Hello,

When viewing a commit, it is hard to know which files are in the commit. It would be nice to use the `--stat` option to display a summary of what files are impacted by the commit.

Something like:
```
git show --stat --patch HEAD
```

cheers

tpope:
You can get a similar effect with

```vim
autocmd User FugitiveCommit setlocal foldmethod=syntax foldlevel=0
```

PR #2155, opened 2023-04-21
Show commit stat to quickly view files of the commit

kawas44:
Hello,

Here is a small PR to use `git show --stat --patch` to quickly see which files are affected by a commit.
I am not really sure that the location of the modification is the right place to do it :grin: 

Related to issue #2154

Cheers


tpope:
The format is stable, for now at least, and closed to bikeshedding. See alternative in #2154.

Issue #2133, opened 2023-03-12
Is it possible: double click -> dv (Gvdiffsplit) the file? 


kgfly:
Currently, after G -> double click the file. It opens the file.

Is it possible do some setting so: double click -> dv (Gvdiffsplit) the file? 

Thanks

tpope:
No setting. It would probably be possible to map.

kgfly:
Thanks, @tpope  How to map that? May you please suggest?

kgfly:
Hi @tpope , may you please share how to do the mapping which only works in G status window?

Thank you very much.

tpope:
Make it a `<buffer>` map, and use the `autocmd User FugitiveIndex` event.

kgfly:
Hi @tpope, I have created a pull request. https://github.com/tpope/vim-fugitive/pull/2160. Do you think the reason is strong enough to justify?

tpope:
No.

kgfly:
Thanks, anyway.

"autocmd User FugitiveIndex nmap <buffer> <2-LeftMouse> dv" work for me.

kgfly:
Hi @tpope , may I ask you one more question: do you know why 

"autocmd User FugitiveIndex nmap <2-LeftMouse> dv" only works on terminal vim, not gui vim (gvim)?

I tried on both Windows and Linux. Neither gvim works with the mapping.

Thanks

PR #2160, opened 2023-04-28
Add fugitive_doubleclick_diff to config double click behavior

kgfly:
People who gets used to VS code like to see the diff by mouse click. The configuration is used to help more people to migrate from VS code to vim/fugitive. Thanks.


tpope:
You can do this without an option:

```vim
autocmd User FugitiveIndex nmap <buffer> <2-LeftMouse> dv
```

kgfly:
Thanks, @tpope 

`autocmd User FugitiveIndex nmap <buffer> <2-LeftMouse> dv`

does not work with gvim. Do you know why? Thanks. 

(I also ask in the thread https://github.com/tpope/vim-fugitive/issues/2133)

PR #2162, opened 2023-04-28
Always fire BufReadStatus's autocmd

### LINKED ISSUE #2161: Staging in Gstatus clobbers index mappings
Related to #2151.

I want to use `s` to toggle stage instead of `-`. I don't care that I lose the stage mapping.

I use the autocmd to define my mapping:

```
autocmd User FugitiveIndex nmap <buffer> s -
```

## Repro

1. `:G` to open status
2. `:nmap s` and see my toggle mapping.
3. use `s` to stage (or `-`)
4. `:nmap s`  and no longer see my toggle mapping. Instead, it's the default fugitive stage mapping.

The output of `let @+ = execute("verb nmap s")` before and after pressing `s`:
```
n  s            @-
	Last set from ~\.vim\bundle\aa-david\reprovimrc.vim line 59

n  s           &@:<C-U>execute <SNR>33_Do('Stage',0)<CR>
	Last set from ~\.vim\bundle\fugitive\autoload\fugitive.vim line 317
```

I assume something in the buffer refresh should be firing the FugitiveIndex autocmd but isn't? However, I can't see where it could be wrong.

## Minimal vimrc
```
" gvim -Nu ~/.vim/bundle/aa-david/reprovimrc.vim -U NONE

let s:plugins = ['sensible']
let s:plugins += ['fugitive']

set runtimepath-=~/.vim
set runtimepath-=~/.vim/after
set runtimepath-=~/vimfiles
set runtimepath-=~/vimfiles/after
for plugin in s:plugins
    exec "set runtimepath^=~/.vim/bundle/". plugin
    exec "set runtimepath+=~/.vim/bundle/". plugin ."/after"
endfor
set viminfofile=NONE

autocmd User FugitiveIndex nmap <buffer> s -
```

Tested on e6651a79facf5cc2b7c554fdc19eb8a9fe89602c. vim 9.0.1425.

idbrii:
Fix #2161.

Always execute BufReadStatus's return value because that's the code that fires the FugitiveIndex autocmd.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -4265,7 +4265,7 @@ function! s:ReloadStatusBuffer(...) abort
   endif
   let original_lnum = a:0 ? a:1 : line('.')
   let info = s:StageInfo(original_lnum)
-  call fugitive#BufReadStatus(0)
+  exe fugitive#BufReadStatus(0)
   call setpos('.', [0, s:StageSeek(info, original_lnum), 1, 0])
   return ''
 endfunction


idbrii:
Worth noting that I don't see any other calls to `fugitive#BufReadStatus` that fail to `:execute`. Same for `fugitive#BufReadCmd` and `s:TempReadPost` which both follow this return autocmd command pattern.

tpope:
The existing behavior is arguably correct. Unlike every other event, this isn't driven by `BufReadCmd`, it's just updating the buffer contents. But defining the maps at the same time does create the problem you reported, so I'll merge for now.

PR #2177, opened 2023-06-01
Docs: Escape `<` character in help.txt

martinbjeldbak:
Escaping issue seem to be causing the `<` character not to be rendered in helptext for the "remove inline diff" command, see screenshots in PR

Using neovim

```
$ nvim --version
NVIM v0.9.0
Build type: Release
LuaJIT 2.1.0-beta3
```

Current `:h fugitive` rendering:
before

After this change:
after

### DIFF
--- a/doc/fugitive.txt
+++ b/doc/fugitive.txt
@@ -302,7 +302,7 @@ X                       Discard the change under the cursor.  This uses
 >                       Insert an inline diff of the file under the cursor.
 
                                                 *fugitive_<*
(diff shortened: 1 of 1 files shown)


tpope:
I'm not seeing this, and your fix causes a double `<` to appear for me. Are you using some nonstandard highlighting, perhaps Tree-sitter?

martinbjeldbak:
Yes you're right! Could be due to tree-sitter and its help syntax highlighting. I will chase it down there. Thank you!

PR #2190, opened 2023-07-06
Update README Install section to include packer.nvim example with link

friendlyantz:
I wan't able to find a link here for installing this, so decided to update README to make life easy for beginners like me. 
let me know. cheers

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -86,7 +86,14 @@ For more information, see `:help fugitive`.
 
 ## Installation
 
-Install using your favorite package manager, or use Vim's built-in package
+Install using your favorite package manager:
+
+i.e [packer.nvim](https://github.com/wbthomason/packer.nvim)
+```lua
+use('tpope/vim-fugitive')
+```
+
+or use Vim's built-in package
 support:
 
     mkdir -p ~/.vim/pack/tpope/start


tpope:
This sounds like packer.nvim is my favorite package manager when in fact I tried it and found it unusably broken. I cannot abide endorsing it in any way.

friendlyantz:
interesting, I didn't realise it was bad - sorry for the bother, very new to the VIM, and I am loving it so far. Will look at using Plug next time. Cheers

tpope:
To be clear, I have no idea if anyone else thinks it's "bad", it's just my personal opinion. And "unusably broken" is quite the allegation so let me elaborate, if not for you then for anyone else that finds this issue.

Lots of Neovim plugins use an idiom where you `require('something').setup({...})` to configure it. I'll use [noice.nvim](https://github.com/folke/noice.nvim) as an example. Say you want to install noice.nvim with packer.nvim and configure it. You might add this to your `init.lua`:

```lua
require('packer').startup(function(use)
  use 'wbthomason/packer.nvim'
  use 'folke/noice.nvim'
end)

require('noice').setup({
   presets = ...
})
```

However, if you try this, you'll get an error:

```
Error detected while processing /home/tpope/.config/nvim/init.lua:                                                                                                                            
E5113: Error while calling lua chunk: /home/tpope/.config/nvim/init.lua:5: module 'noice' not found:                                                                                       
```

That makes sense because Packer hasn't installed it yet. You need to call `:PackerCompile` to do that. But that doesn't mean it isn't a problem. How can we configure the plugin in a way that doesn't break when it isn't installed? Turns out Packer has a solution for this. This is lifted from the [README section on "Specifying plugins"](https://github.com/wbthomason/packer.nvim#specifying-plugins):

```lua
use {
  'myusername/example',        -- The plugin location string
  -- The following keys are all optional
  ...
  config = string or function, -- Specifies code to run after this plugin is loaded.
  ...
}
```

There's several examples of using it elsewhere in that README. Furthermore the noice.nvim README uses this approach in their installation instructions, or at least they did until [it switched to lazy.nvim](https://github.com/folke/noice.nvim/commit/271bd4d0b7c4d9d3df12dfd93d8df42f9bccba69). So perfect, this init.lua should work:

```lua
require('packer').startup(function(use)
  use 'wbthomason/packer.nvim'
  use {'folke/noice.nvim', config = function() require('noice').setup({
    presets = ...
  end}
end)
```

Except ... it doesn't work. And if you look at the [source code](https://github.com/wbthomason/packer.nvim/blob/1d0cf98a561f7fd654c970c49f917d74fafe1530/lua/packer/load.lua#L49-L56) you can plainly see why, it doesn't accept a "string or function", it accepts a list of strings.

Now, I'm not hear to rip on the maintainer for this oversight, it could happen to anyone. But how is this arguably the most popular package manager for Neovim, and yet the first thing I try to do—a thing I would expect to *nearly everyone* to do—completely broken, broken for as long as the plugin has existed (I checked the Git history), and there's not even an issue about it in the issue tracker?

friendlyantz:
ditto! Thank you for a detailed explanation, greatly appreciated

Issue #2217, opened 2023-09-27
[Feature Request] Include commit hash when showing a commit


jbyler:
When displaying a commit object, vim-fugitive calls `git show` with a pretty format that imitates what `git cat-file commit` shows: it includes the tree hash, parent hash, author info, and committer info.

My feature request is to add the commit hash to the pretty format, and ideally also the ref description.

Justification: when viewing info about a commit, I'd like easy access to which commit I'm looking at. I might want to yank the commit hash to use elsewhere, for example.

I'm happy to submit a PR for this if you'll accept it. The PR will do 2 things:
1. add `commit%x20%H%d%n` to the start of the pretty format
2. replace `git cat-file commit <rev>` with a modified `git show` command in the code branch for when `b:fugitive_display_format` is set, so that the format will still match when pressing `a` or `i`

I'm testing this locally and it seems to work fine. I can keep it as a local patch I suppose if you don't think this feature will benefit other users.

richardebeling:
Hey @tpope  what's your position on this? I regularly struggle with this when using `:Git blame`, after pressing enter to look at some commit, I see that it is the commit I'm looking for, but there's no commit hash shown anywhere anymore.

Dr-Willy:
Is there anything blocking the pull request of @jbyler  ? It is really useful to see the commit hash.

jbyler:
I re-tested after resolving conflicts. Still works as desired. I have been using this branch since I created it, and I think it's an important improvement that would benefit all users.

@tpope, would it be possible to get the PR merged?

lf-:
I found myself patching this exact thing because it has also been causing me problems for years and was about to file a PR. In absence of filing my own duplicate PR, I am leaving this comment to note that this is a significant usability problem that I mentioned to my coworker and she was like "yeah this has been driving me nuts for years".

PR #2218, opened 2023-09-27
Issue #2217 Include commit hash and refs in commit display

jbyler:
When displaying a commit, add a new line showing the commit hash and refs description.

Also switches from using `git cat-file commit` to using `git show --no-patch` for the short form toggled with `a` or `i` to keep the format consistent.


Dezzj (commented):
Veiwed

Scutua (approved):

Issue #2219, opened 2023-09-28
Switch from a unified diff into a split diff


idbrii:
When I'm browsing commits, how can I switch from a unified diff (opening the commit in a buffer) to a split diff (Gdiffsplit)?

Since a commit can span multiple files, it'd have to act on the file owning the diff hunks under the cursor.

I can `CR` on a + line to open the file after the change and `:Gdiffsplit !^` to start the diff. Not sure how to do that on - lines.

One of my first guesses was to call `:Gdiffsplit` in the buffer, but that has no effect.

idbrii:
I figured out that the line containing the diff command is a special kind of fugitive-object:

```
diff --git a/data/scripts/file.lua b/data/scripts/file.lua
```

You can `CR` or most of the other `:help fugitive-navigation-maps` keys on that line to open a split diff.

I don't think this is anywhere in the help, so I'll open a PR to add.

PR #2220, opened 2023-09-28
Document opening diff split from patch

idbrii:
References #2219.

There's awesome special behaviour for buffers with diff output to view it in a split diff instead of unified diff.

Let me know if my terminology needs correcting.

I also thought it was great that I could `set foldmethod=syntax foldlevel=0` to collapse diffs down to a list of files that you can press `o` to open in a split diff. But that seems excessively detailed for the doc.


Dezzj (commented):
Thank you

PR #2244, opened 2023-12-15
Fix typo in docs

theletterjeff:


tpope:
This was deliberate. It's attempting to convey "first parent of first parent of first parent of ...".

PR #2258, opened 2024-01-25
Fix misaligned output of :G blame -s

alexklapheke:
When I do `:G blame -s` on a large file, the parentheses after the line numbers are misaligned, leading to ugly output and a too-wide buffer:

blame1

It looks like what's happening is the syntax isn't matching the spaces before the line numbers (highlighted below with `conceallevel=0`), so they aren't getting concealed:

blame2


tpope:
Thanks!

PR #2262, opened 2024-01-30
Fix issue where fugitive resets "last accessed tab" status

lukelbd:
I have experienced a recurring bug when trying to jump to the "last accessed tab" with `:tabnext #` where vim seems to randomly forget the access history and jump to the right-most tab instead. I recently discovered that setting tab-page-scope variables with `settabvar(tabnr, ...)` gives `tabnr` this "last accessed" status, and after grepping `~/.vim`, it seems `vim-fugitive` was causing this behavior by iterating over all tabs and calling `settabvar()` for each one.

This PR preserves "last accessed tab" status by setting the `tabpagenr('#')` tab-scope variable after all the other tab variables. Note that if the "previous tab" is unset then `tabpagenr('#')` returns `0` and `settabvar(...)` fails silently.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -4362,9 +4362,13 @@ function! fugitive#DidChange(...) abort
   if a:0 > 1 ? a:2 : (!a:0 || a:1 isnot# 0)
     let t = reltime()
     let t:fugitive_reload_status = t
+    let prevnr = tabpagenr('#')
     for tabnr in range(1, tabpagenr('$'))
-      call settabvar(tabnr, 'fugitive_reload_status', t)
+      if tabnr != prevnr
+        call settabvar(tabnr, 'fugitive_reload_status', t)
+      endif
     endfor
+    call settabvar(prevnr, 'fugitive_reload_status', t)
     call s:ReloadTabStatus()
   else
     call s:ReloadWinStatus()


Sonyversion8:
> I have experienced a recurring bug when trying to jump to the "last accessed tab" with `:tabnext #` where vim seems to randomly forget the access history and jump to the right-most tab instead. I recently discovered that setting tab-page-scope variables with `settabvar(tabnr, ...)` gives `tabnr` this "last accessed" status, and after grepping `~/.vim`, it seems `vim-fugitive` was causing this behavior by iterating over all tabs and calling `settabvar()` for each one.
> 
> This PR preserves "last accessed tab" status by setting the `tabpagenr('#')` tab-scope variable after all the other tab variables. Note that if the "previous tab" is unset then `tabpagenr('#')` returns `0` and `settabvar(...)` fails silently.

Sonyversion8 (approved):
Access

tpope:
Good catch. That behavior of `settabvar()` is wild.

tpope:
I have added a Vim version check to this.

I've also started getting occasional `ml_get` errors pointing at the new `settabvar()` call. I'm hoping that this is related to the "fails silently" case, as that's easy enough to guard against.

tpope:
> I've also started getting occasional `ml_get` errors pointing at the new `settabvar()` call. I'm hoping that this is related to the "fails silently" case, as that's easy enough to guard against.

Nope, this was tickling a Vim bug introduced in [9.0.0915](https://github.com/vim/vim/commit/29967732761d1ffb5592db5f5aa7036f5b52abf1) and fixed in [9.0.1362](https://github.com/vim/vim/commit/99ad3a8bb95c6f860545a050472b6181e33bac1a). Updated the version constraint accordingly.

justinmk:
> That behavior of `settabvar()` is wild.

Fixed: https://github.com/vim/vim/pull/14017

tpope:
Thanks for the heads up. I think the workaround can stay in place.

PR #2322, opened 2024-07-18
feat: support vim.ui.open for GBrowse on Neovim

stevearc:
It is not uncommon for Neovim users to have netrw disabled in favor of other file explorers. When this is the case, `:GBrowse` fails. Neovim 0.10 ships with a new [vim.ui.open](https://neovim.io/doc/user/lua.html#vim.ui.open()) helper method that is designed to replace this exact functionality.

This PR simply preferentially uses `vim.ui.open` in Neovim 0.10+ over netrw. I have tested it with a simple `:GBrowse` in a file and it appears to work well.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -7433,6 +7433,8 @@ function! s:BrowserOpen(url, mods, echo_copy) abort
     return 'echo '.string(url).'|' . mods . 'Browse '.url
   elseif exists(':OpenBrowser') == 2
     return 'echo '.string(url).'|' . mods . 'OpenBrowser '.url
+  elseif has('nvim-0.10')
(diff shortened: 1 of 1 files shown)


tpope:
Netrw brings a bit more to the table than just opening a URL: It also displays an error message if something went wrong. Despite having `ui` in the name, `vim.ui.open` doesn't concern itself with this UI affordance, or indeed, any other UI at all. We'd need [a dozen lines of Lua](https://github.com/neovim/neovim/blob/27fb62988e922c2739035f477f93cc052a4fee1e/runtime/lua/vim/_defaults.lua#L101-L113) to achieve the same behavior as Netrw, and I don't want to add a `_fugitive.lua` just for that.

As a compromise, I have added support with more limited error handling, but given it a priority _below_ Netrw.

stevearc:
Works for me, thank you!

PR #2353, opened 2024-11-05
Handle range-diff commit header lines in `s:cfile()`

zivarah:
While `range-diff` output doesn't display with any syntax highlighting, it's still in a buffer with a filetype of 'git' and thus fugitive attaches the standard jump mappings. However, those mappings work inconsistently for `range-diff` output due to some accidental matching in `s:cfile()`. For lines like these:

    1:  aaaaaaa = 1:  bbbbbbb My subject
    2:  aaaaaaa = -:  ------- My other subject

Depending on the position of the cursor, we sometimes hit the `<cword>` fallback and open the commit. This can happen when the cursor is over a commit hash, or even on one of the `:`. However, for a line like this:

    -:  ------- > 2:  aaaaaaa My subject

We end up hitting the block intended to catch +/- lines in diff output, and fail to identify a commit.

Add a special set of handling for range-diff header lines. In cases where only one commit is present in the line, open that commit automatically regardless of cursor position. For cases where the line represents a different commit on each side, fall back to a `<cword>` approach to allow opening either commit based on cursor position.


zivarah:
@tpope I wanted to check in on this since it's been a month since I opened it. Let me know if there's anything you'd like to see changed (or if you're concerned about supporting range-diff for some reason). Thanks!

Scutua (approved):

PR #2355, opened 2024-11-11
fix: tab Gdiffsplit compare with empty buffer

kevinhwang91:


tpope:
How do I reproduce the problem?

kevinhwang91:
Run `tab Gdiffsplit` in a file under git repo directly. This worked fine before https://github.com/tpope/vim-fugitive/commit/2a388ef5ba55fd7d50cbefe5767c0e7b65708eef.

tpope:
Strongly suspect I switched from `tab split` to `tabedit` because they interpreted the count differently, but they seem to be behaving the same when I test them now.

PR #2359, opened 2024-12-02
Add support for multi-file edits

### LINKED ISSUE #2352: Edit multiple files / load multiple files into buffer list for `:Git <command>`'s
Assume the command in question is `:Git send-email --annotate HEAD~2`.  This will prepare patches for the last two commits pointed to by `HEAD` and prepare one email message for each which will be sent to the editor for further editing because of the `--annotate` flag.  The default for the option `sendemail.multiEdit` is `true` which will start _one_ editor instance with two buffers loaded (one for each email message) when above command is executed _from the command line_.  The user can then jump back and forth between these buffers and further edit the messages.  When the editor instance is closed `git` will send the emails.  When `sendemail.multiEdit` is `false` the process is sequential, starting a new editor instance for every file that needs editing.

When the above command is executed from within `vim` (with `sendemail.multiEdit = true`) only the first file to edit is loaded into the buffer list.  When that buffer is removed, the command immediately proceeds to sending the email(s), assuming the remaining email messages do not need editing (these buffers are somehow lost in the process).  When `sendemail.multiEdit = false`, the user gets the chance to edit each message but the process is sequential.

Is it possible for a `Git <command>` that expects to edit _multiple_ files, to load all these files into the buffer list at once?  The command must only proceed when all these buffers are removed using something like `:bd /tmp/tmp_folder/*.patch` or some other means.

Thanks,
fabs

fab4100:
Git send-email by default uses multi-file edit mode
(`sendemail.multiEdit`) when more than one patch file requires editing an
associated email message and/or cover-letters.  Multi edit mode spawns
one editor instance with multiple files in argument list.  This behavior
is supported by creating multiple splits for a running vim instance when
`:G send-email` is called.  When the user sets sendemail.multiEdit to
false in the local or global config, if multiple files need editing then
they are processed sequentially in a single split at a time.

Closes #2352

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -3454,9 +3454,9 @@ function! s:RunEdit(state, tmp, job) abort
   endif
   call remove(a:state, 'request')
   let sentinel = a:state.file . '.edit'
-  let files = readfile(sentinel, '')
-  call writefile([len(files)], sentinel)
-  for file in reverse(files)
+  let active_buffers = {}
+  let buf_sequence = []
+  for file in readfile(sentinel, '')
     let file = FugitiveVimPath(file)
     try
       if !&equalalways && a:state.mods !~# '\<\d*tab\>' && 3 > (a:state.mods =~# '\<vert' ? winwidth(0) : winheight(0))
@@ -3473,6 +3473,8 @@ function! s:RunEdit(state, tmp, job) abort
     set bufhidden=wipe
     call s:InitializeBuffer(a:state)
     let bufnr = bufnr('')
+    call add(buf_sequence, bufnr)
+    let active_buffers[bufnr] = file
     let s:edit_jobs[bufnr] = [a:state, a:tmp, a:job, sentinel]
     call fugitive#DidChange(a:state.git_dir)
     if bufnr == bufnr('') && !exists('g:fugitive_event')
@@ -3485,6 +3487,8 @@ function! s:RunEdit(state, tmp, job) abort
       endtry
     endif
   endfor
+  let a:state['active_buffers'] = active_buffers
+  call win_gotoid(bufwinid(buf_sequence[0]))
   return 1
 endfunction
 
@@ -3637,10 +3641,10 @@ if !exists('s:edit_jobs')
   let s:edit_jobs = {}
 endif
 function! s:RunWait(state, tmp, job, ...) abort
+  if has_key(a:state, 'active_buffers')
+    return ''
+  endif
   if a:0 && filereadable(a:1)
-    if a:0 > 1 && a:2 > 0
-      return ''
-    endif
     call delete(a:1)
   endif
   try
@@ -3743,15 +3747,18 @@ function! s:RunBufDelete(bufnr) abort
   endif
   if has_key(s:edit_jobs, a:bufnr) |
     call add(s:resume_queue, remove(s:edit_jobs, a:bufnr))
-    let sentinel = s:resume_queue[-1][0].file . '.edit'
-    let active_buffers = str2nr(readfile(sentinel, '', 1)[0]) - 1
-    call add(s:resume_queue[-1], active_buffers)
-    if active_buffers < 1
-      call feedkeys("\<C-\>\<C-N>:redraw!|call delete(" . string(sentinel) .
-            \ ")|call fugitive#Resume()|checktime\r", 'n')
-    else
-      call writefile([active_buffers], sentinel)
+    let state = s:resume_queue[-1][0]
+    if has_key(state, 'active_buffers')
+        call remove(state.active_buffers, a:bufnr)
+        if !len(state.active_buffers)
+            call remove(state, 'active_buffers')
+        endif
+    endif
+    if has_key(state, 'active_buffers')
       call fugitive#Resume()
+    else
+      call feedkeys("\<C-\>\<C-N>:redraw!|call delete(" . string(state.file . '.edit') .
+            \ ")|call fugitive#Resume()|checktime\r", 'n')
     endif
   endif
 endfunction
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -3454,33 +3454,37 @@ function! s:RunEdit(state, tmp, job) abort
   endif
   call remove(a:state, 'request')
   let sentinel = a:state.file . '.edit'
-  let file = FugitiveVimPath(readfile(sentinel, '', 1)[0])
-  try
-    if !&equalalways && a:state.mods !~# '\<\d*tab\>' && 3 > (a:state.mods =~# '\<vert' ? winwidth(0) : winheight(0))
-      let noequalalways = 1
-      setglobal equalalways
-    endif
-    let mods = s:Mods(a:state.mods, 'SpanOrigin')
-    exe substitute(mods, '\<tab\>', '-tab', 'g') 'keepalt split' s:fnameescape(file)
-  finally
-    if exists('l:noequalalways')
-      setglobal noequalalways
-    endif
-  endtry
-  set bufhidden=wipe
-  call s:InitializeBuffer(a:state)
-  let bufnr = bufnr('')
-  let s:edit_jobs[bufnr] = [a:state, a:tmp, a:job, sentinel]
-  call fugitive#DidChange(a:state.git_dir)
-  if bufnr == bufnr('') && !exists('g:fugitive_event')
+  let files = readfile(sentinel, '')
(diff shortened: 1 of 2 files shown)


Scutua (approved):

Scutua (approved):

fab4100:
To help push progress, I am describing the changes introduced in this PR in more detail below.

1. Multiple file arguments are read by adding a `for`-loop that appends the arguments to the sentinel `$FUGITIVE.edit` (https://github.com/fab4100/vim-fugitive/blob/28f17e54691def71d5f0bf3aa448fe5bd68b70de/autoload/fugitive.vim#L3958)
2. Instead of reading a single file argument, a list of arguments is read (https://github.com/fab4100/vim-fugitive/blob/28f17e54691def71d5f0bf3aa448fe5bd68b70de/autoload/fugitive.vim#L3446) which are then processed in a newly added outer loop (https://github.com/fab4100/vim-fugitive/blob/28f17e54691def71d5f0bf3aa448fe5bd68b70de/autoload/fugitive.vim#L3448), each loop iteration initializes a new buffer with focus on the first file argument given upon loop exit (hence the `reverse(files)` iterator).  Since now multiple buffers may be created, it is no longer sufficient to indicate buffer deletion by simply removing the sentinel (`$FUGITIVE.edit`).  Instead, the number of active buffers is stored into the sentinel to keep track of the number of active buffers when multiple files are being edited (https://github.com/fab4100/vim-fugitive/blob/28f17e54691def71d5f0bf3aa448fe5bd68b70de/autoload/fugitive.vim#L3447).
3. Whenever a buffer is deleted, the active buffer count in the sentinel is decreased https://github.com/fab4100/vim-fugitive/blob/28f17e54691def71d5f0bf3aa448fe5bd68b70de/autoload/fugitive.vim#L3735-L3736.  If the active buffer count is zero, the same code is executed as is currently the case for the single file only edit (deleting the sentinel then resume session https://github.com/fab4100/vim-fugitive/blob/28f17e54691def71d5f0bf3aa448fe5bd68b70de/autoload/fugitive.vim#L3739-L3740).  If one or more buffers are active, the sentinel is updated and the session is resumed without deleting the sentinel (https://github.com/fab4100/vim-fugitive/blob/28f17e54691def71d5f0bf3aa448fe5bd68b70de/autoload/fugitive.vim#L3742-L3743).
4. The active buffer count is appended to the resume queue whose arguments are passed down `s:RunWait` which checks for the newly appended active buffer count argument (see previous item).  If there are 1 or more active buffers, `s:RunWait` returns to caller without taking any action (continuing editing session of active buffers).  If the active buffer count is zero the same code is executed as is currently done for the single file only implementation (which concludes the `$FUGITIVE.edit` session).

Please let me know if I can implement some changes you may desire.  Thanks for taking the time.

autoload/fugitive.vim
```diff
@@ -3443,33 +3443,37 @@ function! s:RunEdit(state, tmp, job) abort
...
-  endtry
-  set bufhidden=wipe
-  call s:InitializeBuffer(a:state)
-  let bufnr = bufnr('')
-  let s:edit_jobs[bufnr] = [a:state, a:tmp, a:job, sentinel]
-  call fugitive#DidChange(a:state.git_dir)
-  if bufnr == bufnr('') && !exists('g:fugitive_event')
+  let files = readfile(sentinel, '')
+  call writefile([len(files)], sentinel)
+  for file in reverse(files)
```
tpope, line 3448: How does this fair with each of `'splitbelow'` and `'nosplitbelow'` set? I would expect one to give an unintuitive order.

It's also a little weird that the buffer numbers would end up backwards. The first file would have the biggest buffer number.

(Perfection isn't mandatory here, given that we currently don't support multiple arguments at all.)

fab4100: `splitbelow` and `nosplitbelow` behave as follows with reference to
currently focused window (see `<W2>` in sketch below):

* `nosplitbelow`: stack grows upwards with first file in focus in top most
split
* `splitbelow`: stack grows downwards with first file in focus in bottom
most split

With reversed sentinel file list iteration the command `:G send-email --annotate
HEAD~2` results in: (created splits are indicated by `S`, numbers indicate
relative buffer number)

```
<  > = focused window                 |  W1  |      |
                                      |------|      |
|  W1  |      |                       | <S2> |      |
|------|      |                       |------|      |
| <W2> |  W4  |  -- nosplitbelow -->  |  S1  |  W4  |
|------|      |                       |------|      |
|  W3  |      |                       |  W2  |      |
                                      |------|      |
                                      |  W3  |      |
```

```
<  > = focused window                 |  W1  |      |
                                      |------|      |
|  W1  |      |                       |  W2  |      |
|------|      |                       |------|      |
| <W2> |  W4  |  --  splitbelow  -->  |  S1  |  W4  |
|------|      |                       |------|      |
|  W3  |      |                       | <S2> |      |
                                      |------|      |
                                      |  W3  |      |
```

The last split is the focused one corresponding to the largest buffer number
holding *the first* file in the sentinel list.  The first file in the sentinel
list corresponds to `[PATCH 1/2]` which is the focused buffer (with the largest
buffer ID) in this case.  Intuitively to me as a user, I would expect that start
editing the first patch and I would be less concerned about the actual buffer ID
relating to this file (while it may be confusing for scripting but I believe `:G
send-email` is mainly interactive).  To me this would be the expected behavior.

Removing the file list reversal would result in the same sketch as above (w/r/t
to buffer IDs) but the active buffer `<S2>` would hold the `[PATCH 2/2]`. This
would be consistent w/r/t buffer ID `<-->` file list in sentinel file, but
*inconsistent* with the file order you are presented when editing the patches
from the terminal with the command `git send-email --annotate HEAD~2` (however,
for this case the buffer ID corresponds to the patch ID in the file as well).

I have pushed a commit for in-memory split buffer book-keeping relating to your
comment below (`89ed871`).  In this commit I have removed the reverse file
iteration but switch focus to the first created buffer instead.  This is
consistent `git send-mail` run from the command line and buffer ID `<-->` patch
ID/sentinel file list.  This results in:

```
<  > = focused window                 |  W1  |      |
                                      |------|      |
|  W1  |      |                       |  S2  |      |
|------|      |                       |------|      |
| <W2> |  W4  |  -- nosplitbelow -->  | <S1> |  W4  |
|------|      |                       |------|      |
|  W3  |      |                       |  W2  |      |
                                      |------|      |
                                      |  W3  |      |
```

```
<  > = focused window                 |  W1  |      |
                                      |------|      |
|  W1  |      |                       |  W2  |      |
|------|      |                       |------|      |
| <W2> |  W4  |  --  splitbelow  -->  | <S1> |  W4  |
|------|      |                       |------|      |
|  W3  |      |                       |  S2  |      |
                                      |------|      |
                                      |  W3  |      |
```

With:

* `S1`: file `0001` (a.k.a `[PATCH 1/2]`)
* `S2`: file `0002` (a.k.a `[PATCH 2/2]`)

autoload/fugitive.vim
```diff
@@ -3725,8 +3732,16 @@ function! s:RunBufDelete(bufnr) abort
...
-    call feedkeys("\<C-\>\<C-N>:redraw!|call delete(" . string(s:resume_queue[-1][0].file . '.edit') .
-          \ ")|call fugitive#Resume()|checktime\r", 'n')
+    let sentinel = s:resume_queue[-1][0].file . '.edit'
+    let active_buffers = str2nr(readfile(sentinel, '', 1)[0]) - 1
+    call add(s:resume_queue[-1], active_buffers)
+    if active_buffers < 1
+      call feedkeys("\<C-\>\<C-N>:redraw!|call delete(" . string(sentinel) .
+            \ ")|call fugitive#Resume()|checktime\r", 'n')
+    else
+      call writefile([active_buffers], sentinel)
```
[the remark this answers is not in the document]
fab4100: I certainly agree with I/O argument.  I have pushed a change that manages active buffers in memory via the `state` dictionary (https://github.com/tpope/vim-fugitive/commit/89ed8713d727ec8c9a980ecd4b98a975b53f09bd).

fab4100:
Hi Tim, I have rebased this branch onto latest `master` and addressed your comment regarding `nosplitbelow` and `splitbelow` here https://github.com/tpope/vim-fugitive/pull/2359#discussion_r2008787581.  Active buffers (possibly multiple) are now handled in memory (contents in the sentinel file are not touched), see my comment here https://github.com/tpope/vim-fugitive/pull/2359#discussion_r2008788198.  Please let me know your thoughts and whether we can proceed.  Thanks!

Issue #2354, opened 2024-11-08
[bug] `'<,'>GBrowse` attaches `%5C` in URL before the `#` character


embe221ed:
after [this commit](https://github.com/neovim/neovim/commit/35e07a24a329fdd7f35640843c9bc983185ca04e), and because of [this line](https://github.com/neovim/neovim/blob/7342e6b00d5e9f67fd5ad4d3fadaf7e501598486/runtime/autoload/netrw.vim#L5395) the `#` character used to point to a specific line/range of lines is escaped, causing the opened URL to look like this: `<path_to_file>%5C#<lines>` which results in 404 Not Found error. Confirmed this by removing the `escape()` call from the mentioned line and re-building neovim

tpope:
A possible fix would be to use `netrw#Open()` if available, and update the escaping to match. That way older versions of Vim can still work with the old behavior of `netrw#BrowseX()`.

embe221ed:
fortunately, most of the escaping in `netrw#BrowseX` doesn't seem to actually be applied to the url in this case. I think that checking if `has('nvim-<version>') && exists('*netrw#Open')` can be sufficient enough.

sluongng:
Im also affected by this as I have a custom browse handler with `#` in the URL https://github.com/sluongng/dotfiles/blob/aae4e38389defc08b7e97745fda9836cb39e2ce5/config/nvim/init.lua#L551-L556

Would love to get the PR merged

PR #2361, opened 2024-12-08
use netrw#open if possible

embe221ed:
issue #2354


tpope:
Do we need the `nvim-0.11` condition? I haven't verified, but I expect this will land in Vim as well if it hasn't already.

urbanautomaton:
Thanks for opening this! To confirm, I'm having this issue in vim 9.1 on macOS:

```
$ vim --version
VIM - Vi IMproved 9.1 (2024 Jan 02, compiled Dec 02 2024 18:58:51)
macOS version - arm64
Included patches: 1-900
Compiled by Homebrew
[features snipped]
```

PR #2369, opened 2025-01-20
Support `git commit --fixup=amend:<commit>` and `--fixup=reword:<commit>` keymaps

KSR-Yasuda:
Besides the current keymap `cf` = `git commit --fixup=<commit>`,
support `--fixup=amend:<commit>` and `--fixup=reword:<commit>` as `cfa` and `cfr`.

Also added a keymap `cff` for the command of `cf`,
since the original keymap `cf` comes to wait for an extra key `a` or `r` till timeout.


tpope:
Please follow the [precedent set by Magit](https://magit.vc/manual/magit/Editing-any-reachable-commit.html).

KSR-Yasuda:
So, you mean this?

Then, the existing binding `cA` changes:
to be used for `commit --edit --fixup=amend:<commit>`;
but now `commit --edit --squash=<commit>`, that ought to be `cn`.

tpope:
> Then, the existing binding `cA` changes:

I guess we should have a grace period that tells users about the change. I'll go ahead and do that.

PR #2373, opened 2025-01-24
Support opting out of deprecated mappings

teoljungberg:
This commit follows d70c42aa503751bde3991be8d1dfc0f1819d211a to remove
the mappings that are deprecated inside of fugitive.vim. For now it is
only `<C-p>` and `<C-n>`. The user opts out by setting
`g:fugitive_legacy_mappings` to 0.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -8012,7 +8012,7 @@ function! fugitive#MapJumps(...) abort
       call s:Map('n', 'O',    ':<C-U>exe <SID>GF("tabedit")<CR>', '<silent>')
       call s:Map('n', 'p',    ':<C-U>exe <SID>GF("pedit")<CR>', '<silent>')
 
-      if !exists('g:fugitive_no_maps')
+      if get(g:, 'fugitive_legacy_mappings', 1) && !exists('g:fugitive_no_maps')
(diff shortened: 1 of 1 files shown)


tpope:
The legacy commands were hurting my tab complete. What are these maps hurting?

teoljungberg:
Muscle memory, I jumped at seeing the deprecation warning and made a autocmd to reset the two deprecated mappings to their builtin equivalents.

tpope:
Muscle memory for the built-ins (equivalent to `j`/`k`)? You use the built-ins in the intended way?

teoljungberg:
It was seeing the deprecation warning itself that irked me, it caused me enough friction to remove it from my configuration with an autocmd to restore `<C-n>` and `-p` to their builtin versions.

We can also argue, and I agree, that having the deprecation warning irk me is serving its purpose of moving to a new kewbinding.

tpope:
> it caused me enough friction to remove it from my configuration with an autocmd to restore `<C-n>` and `-p` to their builtin versions.

This feels like a good solution to your problem actually. It's just 2 maps, so just 2 lines of configuration. And you don't even need autocommands, Fugitive should defer to a global map like `nnoremap <C-n> <C-n>` .

Actually it's more than 2 maps, those are just the 2 that impact your personally. Fundamentally my resistance is I don't want to be responsible for maintaining that going forward.

PR #2378, opened 2025-02-17
`commit --amend` bindings needs `--allow-empty`

KSR-Yasuda:
Shouldn't it set `--allow-empty` option for `commit --amend` bindings such as `ca`,
while `cW` binding or `commit --fixup=reword` has been introduced?

Otherwise, amending such rewording commits by `ca` fails.

* cf. https://github.com/tpope/vim-fugitive/pull/2369


tpope:
> Otherwise, amending such rewording commits by `ca` fails.

The reason `git` requires `--allow-empty` is to make sure empty commits are deliberate. It's impossible to create an empty commit by accident. I'm not going to remove that protection to address one narrow edge case.

KSR-Yasuda:
Okay, thank you.

PR #2389, opened 2025-04-02
Make HEAD lookups work with reftable

### LINKED ISSUE #2371: FugitiveHead() produces wrong value with reftable
Git has a new ref backend called `reftable`.  When that's used, the ref data and reflogs are stored inside a set of binary files and the `.git/HEAD` file always contains `ref: refs/heads/.invalid` (note that `.invalid` is not permitted in a real refname).

As a result, calling `FugitiveHead` returns `.invalid`, which is not correct.  This also causes Airline to reflect the incorrect branch name.

To reproduce this situation, you can set up a repository with Git 2.48 and this shell script:

```
#!/bin/sh

git init -b dev --ref-format=reftable test-repo
cd test-repo
touch abc.txt
git add abc.txt
git commit -m +
```

Once inside, you can open Vim with Fugitive enabled and run `:call FugitiveHead()`.  The correct result is `dev`, and the incorrect result is `.invalid`.

I suspect you are going to need to actually invoke Git in this case.  Perhaps a command like `git rev-parse --abbrev-ref HEAD` will be useful.

bk2204:
Git has a new ref storage backend called reftable that stores all refs, including HEAD, as well as all reflogs, in a binary format under .git/reftable.  Because the HEAD file is important in determining whether a directory is a Git repository, Git retains this file, but it always contains "ref: refs/heads/.invalid", thus pointing to an invalid ref, since ref components may not start with a dot.

In such a configuration, the only practical possibility is to invoke a Git command to resolve HEAD for us, so use git rev-parse to do so.  Look up the object ID and the value for HEAD at the same time to avoid the overhead of two calls and use the former if the latter is HEAD (that is, we're not on a branch).

Unfortunately, this doesn't work when we have an unborn branch without any commits, such as when a repository is newly initialized.  Fall back to git symbolic-ref in such a case.

Note that the head cache cannot be preserved here because the .git/HEAD file does not have to change when the branch changes with reftable (in fact, it will not), so remove that code here.

Fixes #2371

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -1016,6 +1016,7 @@ function! s:StdoutToFile(out, cmd, ...) abort
 endfunction
 
 let s:head_cache = {}
+let s:head_reftable_cache = {}
 
 function! fugitive#Head(...) abort
   let dir = a:0 > 1 ? a:2 : s:Dir()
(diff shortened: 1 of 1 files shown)


bk2204:
We could cache on `.git/reftable/tables.list` for reftable plus always `.git/HEAD`.  The former is the list of reftable files, so that would reflect the state of the refs database for reftable.

I can add that functionality, but it may take me a day or two to get it done.

tpope:
I would encourage you to use the command conditionally once `.invalid` is encountered, rather than replacing the existing implementation.

bk2204:
Okay, I've updated this to keep the regular implemntation and only use the reftable implementation only when `refs/heads/.invalid` is detected.  In addition, there's now a cache for the reftable file as well to make this cheaper when the branch doesn't change.

Bhadboiab (approved):
Gratitude

benknoble:
Hi there; anything I can help do to push this forward? I'm starting to convert more of my repos to reftables and noticed this again. Having `.invalid` show up in the status line was more than a little confusing until I retraced memory lane.

PR #2439, opened 2025-10-21
call BrowseX with 1 arg for patch >= 9.1.1485

Ja-sonYun:
https://github.com/vim/vim/commit/ef925556cbea4d54e73388da3b3e6c06a1bbcc02

patch version checked with `git tag --contains ef92555` in vim repository.

`netrw#BrowseX()` was changed in Vim patch 9.1.1485 to take a single argument and handle remote detection internally. Use one-arg call when that patch is present, and fallback to the old two-arg form for older builds.

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -7453,7 +7453,11 @@ function! s:BrowserOpen(url, mods, echo_copy) abort
     elseif exists('*netrw#os#Open')
       return 'echo '.string(url).'|' . mods . 'call netrw#os#Open('.string(url).')'
     elseif exists('*netrw#BrowseX')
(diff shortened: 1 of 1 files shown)


tpope:
Would it be feasible to check the version in `g:netrwPlugin_loaded` instead?

Ja-sonYun:
Do you mean g:loaded_netrwPlugin? If so, that’s what I initially considered too.
However, in patch 9.1.1486 the value of g:netrwPlugin_loaded is still v181, so users building Vim from that patch will get an incorrect function usage.
Additionally, patch 9.1.1484 also v181, so the variable does not accurately indicate the behavioral change introduced in 9.1.1485.

Issue #2448, opened 2026-01-16
Including --stat in fugitive commit view


tonystr:
Hello, when working with git in the terminal, I like to use `git show -p --stat <rev>`, which gives a helpful overview over files changed[0] before the diffs for the commit. 

I initially tried to replicate this with `:Gedit --stat <rev>`, but just got an empty buffer. Looking into the code, I see that the git command for `git show ... <rev>` is hard coded, with a very specific --pretty=format:... (presumably for highlighting), so understandably it would be difficult to allow arbitrary flags to `git show` through `:Gedit`.

Neogit shows this --stat output by default in their commit view[1]. I hacked it in locally to see how it would look in fugitive[1].

If interesting to others, I could implement this with highlighting either as standard or as an option (I believe the code supports ++flags), otherwise I might just fork for my self.

[0] - `git show -p --stat <rev>` output:
<img width="1477" height="1044" alt="Image" src="https://github.com/user-attachments/assets/fcbd739c-db2f-4d89-b27c-2c7ceb927ea5" />

[1] - vim-fugitive :Gedit modified to include `-p --stat` on the left, and Neogit's :NeogitCommit on the right:
<img width="2507" height="1526" alt="Image" src="https://github.com/user-attachments/assets/ece03424-976c-447f-ab79-c9d759dc5a24" />

PR #2449, opened 2026-01-17
feat: add --stat output to commit view as 0th display format

tonystr:
Solves #2448 

This PR adds a third fugitive_display_format to which simply includes `--stat` output in the commit message. Being the 0th display format (handles in the else, rather than an elseif), means that this format will be the default format shown when viewing commits (`:Gedit <rev>`). See appendix[0] for an image of how this looks.

PROBLEMS
   1. - [ ] The `--stat` output is not highlighted by the git parser for treesitter. This would require a PR to the treesitter git parser to solve.
2. - [ ] Display format doesn't appear to be documented in fugitive's docs, this should be added.
3. - [ ] Display format might not be the right way to handle this. Presumably, most users would prefer to always see `--stat` output or to never see `--stat` output. This should probably instead be handled by a config option.
4. - [ ] If display format is to be kept, it should probably not be limited to a given range, rather than be represented as an arbitrary number. 

APPENDIX
[0] - Fugitive `:Gedit` with `--stat` on the left, compared with Neogit's `:NeogitCommit` on the right
536990632-ece03424-976c-447f-ab79-c9d759dc5a24

### DIFF
--- a/autoload/fugitive.vim
+++ b/autoload/fugitive.vim
@@ -3228,11 +3228,26 @@ function! fugitive#BufReadCmd(...) abort
           call s:ReplaceCmd([dir, 'cat-file', '-p', rev])
         endif
       elseif b:fugitive_type ==# 'commit'
-        let b:fugitive_display_format = b:fugitive_display_format % 2
-        if b:fugitive_display_format
+        let b:fugitive_display_format = b:fugitive_display_format % 3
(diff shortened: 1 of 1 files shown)


brighamcampbell:
I'd like to have this kind of functionality! I find the diffstat to be a useful summary.

PR #2489, opened 2026-08-20
Jump from staged inline hunks to work tree

pedro757:
## Problem

In the `:Git` status window, pressing `=` expands a file into an inline diff. Pressing `<CR>` on an unstaged hunk opens the real work tree file at the corresponding line. The same action on a staged hunk instead opens an index or `HEAD` blob, producing a `fugitive://...//0/...` buffer rather than the editable file.

This makes identical navigation behave differently between the Unstaged and Staged sections. It also means that using the staged diff as a way to inspect a change and jump into the file unexpectedly leaves the user in a virtual Fugitive buffer.

Changing only the target path is not enough. Staged hunk positions refer to `HEAD` or the index, while the destination is the work tree. Additional unstaged edits may insert or remove lines before the staged change, so the original offset can point at the wrong line.

## Change

- open the work tree file when jumping from an expanded staged hunk, provided that file still exists
- translate index positions through subsequent unstaged changes before opening the file
- map staged replacements through the index, while mapping pure deletions directly from `HEAD` so restored lines are found correctly
- force zero inter-hunk context for translation diffs so `diff.interHunkContext` cannot merge hunks and distort line mapping
- retain the existing revision-buffer fallback when the file has been deleted from the work tree

## Testing

- exercised additions, context lines, pure deletions, restored deletions, replacements, multi-line replacements, renames, and paths containing spaces in headless Neovim
- verified partially staged files with line-shifting unstaged edits and `diff.interHunkContext=100`
- verified forward and reverse revision line translation
- ran `git diff --check`


zivarah:
I'm relatively certain that the current behavior is very much intended and is a feature, not a bug, and that this would be an undesired breaking change. Being able to edit the staged version directly is a neat feature and (in my opinion) `<CR>` taking you there is appropriate: that's the source location for the diff line that you interacted with.

FYI, you can just use `Gedit` (which I use frequently enough to have bound to `<leader>ge`) to flip from the staged virtual buffer to the real worktree file and vice versa. `Gedit` when viewing a file at some revision also takes you to the worktree file.

I'm not a maintainer here, but I suspect this PR should just be closed.

